PR #2, opened 2020-12-07
Fix README typos

rmccorm4:
Signed-off-by: Ryan McCormick <rmccormick@nvidia.com>


klueska:
Thanks for the PR.

I actually found a couple of other typos while looking at this. Do you want to just incorporate these into your PR? The diff below is *after*  I have applied your changes:
```
$ git diff
diff --git a/README.md b/README.md
index a1aab0b..fce760f 100644
--- a/README.md
+++ b/README.md
@@ -226,11 +226,11 @@ There are two top-level directories in this repository:
 * `/gen`
 * `/pkg`

-The `/gen` directory is uses to house any code used in the _generation_ of the
+The `/gen` directory is used to house any code used in the _generation_ of the
 final Go bindings. The `/pkg` directory is used to house any static packages
 associated with this project as well as the actual Go bindings once they have
 been generated. The one exception is the code used to dynamically load the
-`libnvidia-ml.so` code from a host system attaches the go bindings to it. This
+`libnvidia-ml.so` code from a host system and attach the go bindings to it. This
 package requires no generated code and is housed statically under `pkg/dl`.
 Once the code under `gen/nvml` has passed through `c-for-go` and any manual
 manual wrappers applied, the final generated bindings are placed under
@@ -354,7 +354,7 @@ the end user.

 The `cgo_helpers.go` file defines functions that help in dealing with the types
 coming out of the C API and turning them into more usable Go types. It is
-actually a stripped down version of the autogenerated `cgo_helpers.go` file
+actually a stripped down version of the auto-generated `cgo_helpers.go` file
 from `c-for-go` that we have whittled down to the bare essentials. We also
 define a few of our own functions in here as well. For example, doing things
 like finding the length of a `NULL` terminated string inside a byte slice
@@ -363,7 +363,7 @@ like finding the length of a `NULL` terminated string inside a byte slice

 The `return.go` file simply wraps the `Return` type created by `c-for-go`
 (which is a go-ified version of the `nvmlReturn_t` type from C) and has it
-implement the `Error` interface so it an be returned as a normal Go `error`
+implement the `Error` interface so it can be returned as a normal Go `error`
 type if desired. The string returned as part of the error is the result of
 calling `nvmlErrorString()` under the hood.

@@ -548,12 +548,12 @@ new types are added, or a new API is created that does something outside the
 scope of what has been done so far. These guidelines should be good a starting
 point though.

-Keep in mind, that all updates to the nvml bindings code should be made in the
+Keep in mind, that all updates to the NVML bindings code should be made in the
 `gen/` directory of the repository. Only when releasing new bindings will this
 code be processed and pushed to the `pkg/` directory for release.

 ## Releasing
-Once the coode in `gen/` has been fully updated to support a particular version of NVML, a new release should be created.
+Once the code in `gen/` has been fully updated to support a particular version of NVML, a new release should be created.

 As part of the release, two things need to happen:
 1. A new set of bindings needs to be generated from the code under `gen/` and committed into `pkg/`
```

klueska:
Please just squash them into your original ccommit

rmccorm4:
Done 👍

klueska (approved):
Looks great. Thanks for submitting!

PR #4, opened 2021-01-14
Add docker build

elezar:
This change adds a basic Dockerfile that includes the c-for-go
dependency to allow the generation of bindings without installing
this dependency.

A generic make target `docker-%` is added to run any of the recognised
make targets in the container.

```bash
$ make \t\t
all                    clean                  docker-all             docker-clean           docker-test            image                  test-bindings
bindings               clean-bindings         docker-bindings        docker-clean-bindings  docker-test-bindings   test
```

Note that this pins:
* The GO version to 1.14.4
* The c-for-go commit to 8eeee8c3b71f9c3c90c4a73db54ed08b0bba971d

To ensure that this does not introduce any changes when generating
the bindings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-GEN_DIR = $(PWD)/gen
-PKG_DIR = $(PWD)/pkg
-GEN_BINDINGS_DIR = $(GEN_DIR)/nvml
-PKG_BINDINGS_DIR = $(PKG_DIR)/nvml
+DOCKER ?= docker
+GOLANG_VERSION ?= 1.14.4
+C_FOR_GO_TAG ?= 8eeee8c3b71f9c3c90c4a73db54ed08b0bba971d
+BUILDIMAGE ?= nvidia/go-nvml-devel:$(GOLANG_VERSION)-$(C_FOR_GO_TAG)
+
+PWD := $(shell pwd)
+GEN_DIR := $(PWD)/gen
+PKG_DIR := $(PWD)/pkg
+GEN_BINDINGS_DIR := $(GEN_DIR)/nvml
+PKG_BINDINGS_DIR := $(PKG_DIR)/nvml
 
 SOURCES = $(shell find $(GEN_BINDINGS_DIR) -type f)
 
-.PHONY: all test clean
-.PHONY: bindings test-bindings clean-bindings
+TARGETS := all test clean bindings test-bindings clean-bindings
+DOCKER_TARGETS := $(patsubst %, docker-%, $(TARGETS))
+.PHONY: $(TARGETS) $(DOCKER_TARGETS)
+
+.DEFAULT_GOAL = all
 
 all: bindings
 test: test-bindings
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -46,3 +49,24 @@ test-bindings: $(PKG_BINDINGS_DIR)
...
+docker-%: image
+	@echo "Running 'make $(*)' in docker container $(BUILDIMAGE)"
+	$(DOCKER) run \
+		--rm \
+		-e PWD \
+		-v $(PWD):$(PWD) \
+		-w $(PWD) \
+	    $(BUILDIMAGE) \
+			make $(*)
+
```
klueska, line 69: The spacing here looks strange

Makefile
```diff
@@ -46,3 +49,24 @@ test-bindings: $(PKG_BINDINGS_DIR)
...
 	rm -rf $(PKG_BINDINGS_DIR)
+
+image: Dockerfile
+	DOCKER_BUILDKIT=1 \
+	$(DOCKER) build \
+	    --progress=plain \
+	    --tag $(BUILDIMAGE) \
+	    .
+
+docker-%: image
```
klueska, line 60: Can you add specific targets for the `docker` variants so that they exist for tab-complete?

It should be as simple as changing this to:
```
TARGETS = all test clean bindings test-bindings clean-bindings
docker-$(TARGETS): image
	...
```
We can probably then change the .PHONY line above to just reference $(TARGETS) as well.

Makefile
```diff
@@ -46,3 +49,24 @@ test-bindings: $(PKG_BINDINGS_DIR)
...
+		--rm \
+		-e PWD \
+		-v $(PWD):$(PWD) \
+		-w $(PWD) \
+	    $(BUILDIMAGE) \
+			make $(*)
+
+# Override the target for removing the bindings directory recursively from the container
+docker-clean-bindings: clean-bindings
+	@echo "Skipping containerized removal"
```
klueska, line 72: Why do we want to override this and not do it from docker?

Makefile
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+DOCKER ?= docker
+BUILDIMAGE ?= go-nvml-build
```
klueska, line 16: can we prefix this with `nvidia/`

Makefile
elezar: May be a tabs vs spaces thing. Will check.

Makefile
elezar: Since we're mapping in `$(pwd)` I didn't want to inadvertently delete `/` due to an unset variable.

Makefile
elezar: Sure.

Makefile
elezar: (Yes, was spaces instead of tabs).

Makefile
elezar: Updated:
```
$ make\t\t
all                    clean                  docker-all             docker-clean           docker-test            image                  test-bindings
bindings               clean-bindings         docker-bindings        docker-clean-bindings  docker-test-bindings   test
```

Makefile
```diff
@@ -12,15 +12,23 @@
...
 PKG_BINDINGS_DIR = $(PKG_DIR)/nvml
 
 SOURCES = $(shell find $(GEN_BINDINGS_DIR) -type f)
 
-.PHONY: all test clean
-.PHONY: bindings test-bindings clean-bindings
+.PHONY: $(TARGETS)
+TARGETS := $(NATIVE_TARGETS) $(DOCKER_TARGETS)
+NATIVE_TARGETS := all test clean bindings test-bindings clean-bindings
+DOCKER_TARGETS := $(patsubst %, docker-%, $(NATIVE_TARGETS))
```
klueska, line 29: Does it work to define `.PHONY` before defining `TARGETS`? Also, does it work to define `TARGETS := $(NATIVE_TARGETS) $(DOCKER_TARGETS)` before `NATIVE_TARGETS` and `DOCKER_TARGETS`.  It could see how it would work if it was a `=` instead of a `:=`, but I thought `:=` actually evaluated as it was encountered (not lazily like normal `=`).

Makefile
```diff
@@ -12,15 +12,23 @@
...
+
+$(DOCKER_TARGETS): %: docker-%
```
klueska, line 31: It seems unnecessary to have this.

I think the target below for `docker-%` can just be changed to `$(DOCKER_TARGETS)`.

Makefile
```diff
@@ -46,3 +54,27 @@ test-bindings: $(PKG_BINDINGS_DIR)
 
 clean-bindings:
 	rm -rf $(PKG_BINDINGS_DIR)
+
+# Generate an image for containerized builds
+# Note: This image is local only
+image: Dockerfile
```
klueska, line 60: Can we call this target `.image` instead of just `image` so that it is hidden from the tab completion?

Makefile
elezar: Let me see. It's been a while since I've had to write real make files, so I could definitely improve things.

Makefile
elezar: I can do that.

Makefile
```diff
@@ -46,3 +54,27 @@ test-bindings: $(PKG_BINDINGS_DIR)
...
+	DOCKER_BUILDKIT=1 \
```
klueska, line 61: What is the reason for running with `BUILDKIT`? Is this available for everyone or only if you have `experimental` set up in your `daemon.josn` file?

Makefile
elezar: You're right. Updating so that `.PHONY:` works as expected:

```
$ make -pq : | grep -E "^\.PHONY"
make: *** No rule to make target ':'.  Stop.
.PHONY: all test clean bindings test-bindings clean-bindings docker-all docker-test docker-clean docker-bindings docker-test-bindings docker-clean-bindings
```

elezar: If `.PHONY` is set before `TARGETS` (even with immediate assignment), this does not work.

Makefile
elezar: I had no specific reason. I was checking what `container-runtime` does. Since this is experimental, I will remove it.

elezar: (I did hear that it was faster)

klueska:
Looks great now. Thanks.

PR #5, opened 2021-01-14
Upgrade to nvml 11.1

elezar:
This also adds changes to the Makefile that make updating the bindings simpler in future.


gen/nvml/device.go
```diff
@@ -101,14 +101,14 @@ func (Device Device) GetSerial() (string, Return) {
 }
 
 // nvml.DeviceGetCpuAffinity()
-func DeviceGetCpuAffinity(Device Device, NumCPUs int) ([]uint, Return) {
+func DeviceGetCpuAffinity(Device Device, NumCPUs int) ([]uint32, Return) {
```
elezar, line 104: The interface has remained as `unsigned long *` for the C definition, and this change from `uint` to `uint32` may be due to my platform. I will check up on how to better handle the `long` translation here tomorrow.

gen/nvml/init.go
```diff
@@ -88,6 +88,9 @@ var nvmlDeviceGetNvLinkRemotePciInfo = nvmlDeviceGetNvLinkRemotePciInfo_v1
 var nvmlDeviceRemoveGpu = nvmlDeviceRemoveGpu_v1
 var nvmlDeviceGetGridLicensableFeatures = nvmlDeviceGetGridLicensableFeatures_v1
 var nvmlEventSetWait = nvmlEventSetWait_v1
+var nvmlDeviceGetAttributes = nvmlDeviceGetAttributes_v1
```
elezar, line 91: I think we could generate this and the code for `updateVersionedSymbols()` from the header file directly. The pattern is quite straight-forward.

gen/nvml/device.go
elezar: This was due to the version of `c-for-go` that was update when building the Docker image I used to generate the bindings.

gen/nvml/init.go
elezar: Not critical for this review though.

.gitignore
```diff
@@ -1,3 +1,2 @@
 *.swp
 *.swo
-pkg/nvml/
```
klueska, line 3: I think we should leave this in here. We should be explicit about when we want to update this folder, and you can do it with an explicit `git add -f pkg/nvml`

gen/nvml/nvml.h.patch
```diff
@@ -1,56 +1,61 @@
...
   */
  #define NVML_VALUE_NOT_AVAILABLE (-1)
  
 -typedef struct nvmlDevice_st* nvmlDevice_t;
-+typedef struct {
-+	struct nvmlDevice_st* handle;
++typedef struct
++{
++    struct nvmlDevice_st* handle;
 +} nvmlDevice_t;
```
klueska, line 11: Why did this change? Looks like maybe it was updated to adhere to the consistent format for struct in the rest of this file (which is a good change, actually if that's the reason).

Ditto to similar ones below.

.gitignore
elezar: I was just thinking that there was at least one other situation where we have forgotten to add generated files. Is there a way that we can check this "automatically"?

I don't feel too strongly about it though and I'm perfectly happy to revert this line.

pkg/nvml/nvml.go
```diff
@@ -27,37 +27,37 @@ package nvml
 import "C"
 import "unsafe"
 
-// nvmlInit_v2 function as declared in nvml/nvml.h:1732
+// nvmlInit_v2 function as declared in nvml/nvml.h:1766
```
klueska, line 30: I wonder if there is an option in `c-for-go` to omit this info. It would make looking at the diffs much easier.

klueska (commented):
Looks great overall. I think in the future it may be useful to split this into 4 commits -- one for each of the steps in https://github.com/NVIDIA/go-nvml#updating-the-code. It would have made it slightly easier to review and verify that everything was done from each of those steps.

.gitignore
klueska: Yeah, I think that was a one-of case. In general, I picture us more often making changes where we explicitly *don't* want to commit back to `pkg` until the commit / PR where we do the release.

elezar:
> Looks great overall. I think in the future it may be useful to split this into 4 commits -- one for each of the steps in https://github.com/NVIDIA/go-nvml#updating-the-code. It would have made it slightly easier to review and verify that everything was done from each of those steps.

I can look at doing that for this PR too. One issue that I see with that is that the commits then don't represent functional units, but since we're using merge commits to `master` this probably doesn't make too much difference.

pkg/nvml/nvml.go
elezar: It would seem as if this is [hard-coded](https://github.com/xlab/c-for-go/blob/73cdccf1de24f229fe86ae09fb05977ffc50dabf/generator/gen_const.go#L22), but we could strip these from the code in `pkg` with a `sed` command or something before adding?

An alternative would be a PR to `c-for-go` that adds this optionally.

gen/nvml/nvml.h.patch
elezar: When looking at the input file, the convention used seems to be to start the block on the line following the typedef. For this reason, I constructed the `sed` command to generate the `patch` file to also break after the `typedef` line to ensure uniformity (as you assumed).

From the README:
```bash
sed -E 's#(typedef\s+struct)\s+(nvml.*_st\*)\s+(nvml.*_t);#\1\n{\n    struct \2 handle;\n} \3;#g' gen/nvml/nvml.h \
    | diff -u gen/nvml/nvml.h - > gen/nvml/nvml.h.patch
```

.gitignore
elezar: > Yeah, I think that was a one-of case. In general, I picture us more often making changes where we explicitly _don't_ want to commit back to `pkg` until the commit / PR where we do the release.

OK. This PR now contains this commit, but I can drop it if required.

gen/nvml/nvml.h.patch
elezar: As a note, I have added a make target that updates the file instead of using a patch file. We can revisit if there are additional changes that need to be made to the file that are not purely structural.

README.md
```diff
@@ -292,15 +292,18 @@ symbols include the following (as defined in `nvml.h`):
...
+    #define nvmlDeviceGetCount                      nvmlDeviceGetCount_v2
+    #define nvmlDeviceGetHandleByIndex              nvmlDeviceGetHandleByIndex_v2
+    #define nvmlDeviceGetHandleByPciBusId           nvmlDeviceGetHandleByPciBusId_v2
+    #define nvmlDeviceGetNvLinkRemotePciInfo        nvmlDeviceGetNvLinkRemotePciInfo_v2
+    #define nvmlDeviceRemoveGpu                     nvmlDeviceRemoveGpu_v2
+    #define nvmlDeviceGetGridLicensableFeatures     nvmlDeviceGetGridLicensableFeatures_v3
+    #define nvmlEventSetWait                        nvmlEventSetWait_v2
+    #define nvmlDeviceGetAttributes                 nvmlDeviceGetAttributes_v2
+    #define nvmlDeviceGetComputeRunningProcesses    nvmlDeviceGetComputeRunningProcesses_v2
+    #define nvmlDeviceGetGraphicsRunningProcesses   nvmlDeviceGetGraphicsRunningProcesses_v2
```
klueska, line 306: I would technically expect this in the next commit, but it's OK here too.

klueska (approved):
This looks great. Thanks for doing all the "meta" stuff around automating some of the steps and updating the README. Only one small comment which is just a technicality :)

README.md
elezar: Updated.

elezar:
@klueska since I rebased with the changes to the README in the correct commit, do you mind just giving this the once-over again?

PR #6, opened 2021-01-19
Update Return type with an exrtra API call and fix bug in versioned APIs

klueska:
```
Author: Kevin Klues <kklues@nvidia.com>
Date:   Tue Jan 19 16:28:58 2021 +0000

    Fix error check in call to set versioned API calls

    Previously, it was checking them against nvml.Return types (which was
    wrong on many levels). It only happened wot work on my tests because I
    had a machine with the latest API calls for everything.

    Signed-off-by: Kevin Klues <kklues@nvidia.com>
```
```
Author: Kevin Klues <kklues@nvidia.com>
Date:   Tue Jan 19 16:27:06 2021 +0000

    Add Value() call to nvml.Return type

    Not necessary when using the package API directly, but helps with
    mocking out the Return type in higher level libraries and then having
    them be able to easily swap in the package based implementation at a
    later time.

    Signed-off-by: Kevin Klues <kklues@nvidia.com>
```

### DIFF
--- a/gen/nvml/return.go
+++ b/gen/nvml/return.go
@@ -19,6 +19,10 @@ func ErrorString(Result Return) string {
 	return nvmlErrorString(Result)
 }
 
+func (r Return) Value() Return {
+	return r
+}
+
 func (r Return) String() string {
 	return ErrorString(r)
 }
--- a/gen/nvml/init.go
+++ b/gen/nvml/init.go
@@ -91,50 +91,50 @@ var nvmlEventSetWait = nvmlEventSetWait_v1
 
 // updateVersionedSymbols()
 func updateVersionedSymbols() {
-	ret := nvml.Lookup("nvmlInit_v2")
-	if ret == SUCCESS {
+	err := nvml.Lookup("nvmlInit_v2")
+	if err == nil {
 		nvmlInit = nvmlInit_v2
 	}
-	ret = nvml.Lookup("nvmlDeviceGetPciInfo_v2")
-	if ret == SUCCESS {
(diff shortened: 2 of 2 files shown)


gen/nvml/return.go
```diff
@@ -19,6 +19,10 @@ func ErrorString(Result Return) string {
 	return nvmlErrorString(Result)
 }
 
+func (r Return) Value() Return {
```
elezar, line 22: Just to be clear, this isn't used at the moment?

elezar (commented):
Thanks @klueska. Do we have any idea how to test these going forward?

gen/nvml/return.go
klueska: It's not used anywhere in this repo, correct.
But I am using it in a wrapper I am building.

klueska:
> Thanks @klueska. Do we have any idea how to test these going forward?

To properly test this, we would need machines that supported all of the APIs.
That said (barring that) I think it would be worthwhile to set some CI infrastructure on GitLab to smoke test what we can on a subset of the APIs.  There is a minimal test file you can run under `pkg/nvml` that I run before submitting and PRs to make sure that things aren't too broken.

elezar:
> That said (barring that) I think it would be worthwhile to set some CI infrastructure on GitLab to smoke test what we can on a subset of the APIs. There is a minimal test file you can run under pkg/nvml that I run before submitting and PRs to make sure that things aren't too broken.

Ok. Sounds good. Let's track adding that once we have the repo set up in GitLab (that is already on my TODO list).

PR #9, opened 2021-02-03
Remove funcs attached to Return type

klueska:
Having these here makes it hard to mock this type, and turning it into
    an interface (at this level) is hard because then you can't properly
    compare it to the const definition for SUCCESS, NOT_SUPPORTED, etc.
    (without a cast).

    Until we come up with a better way to support mocking this type with
    funcs attached to it, it's better to just have the top level function
    defined (which will never change / go away.


elezar (approved):
The changes seem to include only some of the modified `pkg` files, and I assume that no others were changed by this since we're not using the return value as something that implements the `Error` interface anywhere?

klueska:
Other than `return.go`, the only other files that are part of the "generated" ones are those where the timestamps change from c-for-go. None of our internal code uses the `Return` type directly.

Issue #11, opened 2021-02-28
GetProcessUtilization says: Insufficient Size


qisikai:
![image](https://user-images.githubusercontent.com/14859076/109418194-0659be00-7a02-11eb-8d03-10a965ea7ef7.png)

I run pytorch in a docker container, with the 4th gpu.
I run go-nvml in host environment. 

Code:
```
_, ret := dev.GetProcessUtilization(ts)
if ret != nvml.SUCCESS {
	log.Printf("[x] Unable to call GetProcessUtilization", nvml.ErrorString(ret))
}

```

OUTPUTS:
```
   Unable to call GetProcessUtilization%!(EXTRA string=Insufficient Size)
```

qisikai:
I modify the code, and this code works:
```
var ProcessSamplesCount uint32 = 100
Utilization := make([]ProcessUtilizationSample, 100)
ret := nvmlDeviceGetProcessUtilization(Device, &Utilization[0], &ProcessSamplesCount, LastSeenTimeStamp)
```

elezar:
Thanks @qisikai. It seems as if the first call to `nvmlDeviceGetProcessUtilization` which should be setting the number of samples is not returning the expected value in `ProcessSamplesCount`. Could you check what value `ProcessSamplesCount` after the first call?

As a matter of interest, what is the timestamp value that you pass?

qisikai:
@elezar 
## 1
value: `ProcessSamplesCount =100` （same value on all the gpus, It's the same with or without load).

## 2
I have tried:  `timestamp=0`, `timestamp = current second * 1000`, `timestamp = (current second - 10s) * 1000 microseconds`,
But none of them work

## 3
I find :
1. when calling `ret := nvmlDeviceGetProcessUtilization(Device, &Utilization[0], &ProcessSamplesCount, LastSeenTimeStamp)`, `ProcessSamplesCount` must be `100`,  `0` or `10` don't work.

2. `&Utilization[0]` must not be nil.


## 4 
Follow code works:
```
// nvml.DeviceGetProcessUtilization()
func DeviceGetProcessUtilization(Device Device, LastSeenTimeStamp uint64) ([]ProcessUtilizationSample, Return) {
	var ProcessSamplesCount uint32 = 100
	Utilization := make([]ProcessUtilizationSample, 100)

	ret := nvmlDeviceGetProcessUtilization(Device, &Utilization[0], &ProcessSamplesCount, LastSeenTimeStamp)
[...]

PR #12, opened 2021-03-01
Check for ERROR_INSUFFICIENT_SIZE for GetProcessUtilization

elezar:
The initial call to nvmlDeviceGetProcessUtilization to determine the number
of samples for which a buffer must be allocated returns ERROR_INSUFFICIENT_SIZE
and not SUCCESS.

Addresses #11 

Signed-off-by: Evan Lezar <elezar@nvidia.com>

### DIFF
--- a/gen/nvml/device.go
+++ b/gen/nvml/device.go
@@ -1468,7 +1468,7 @@ func (Device Device) GetGridLicensableFeatures() (GridLicensableFeatures, Return
 func DeviceGetProcessUtilization(Device Device, LastSeenTimeStamp uint64) ([]ProcessUtilizationSample, Return) {
 	var ProcessSamplesCount uint32
 	ret := nvmlDeviceGetProcessUtilization(Device, nil, &ProcessSamplesCount, LastSeenTimeStamp)
-	if ret != SUCCESS {
+	if ret != ERROR_INSUFFICIENT_SIZE {
 		return nil, ret
 	}
 	if ProcessSamplesCount == 0 {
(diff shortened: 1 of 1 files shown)


gen/nvml/device.go
```diff
@@ -1468,7 +1468,7 @@ func (Device Device) GetGridLicensableFeatures() (GridLicensableFeatures, Return
 func DeviceGetProcessUtilization(Device Device, LastSeenTimeStamp uint64) ([]ProcessUtilizationSample, Return) {
 	var ProcessSamplesCount uint32
 	ret := nvmlDeviceGetProcessUtilization(Device, nil, &ProcessSamplesCount, LastSeenTimeStamp)
-	if ret != SUCCESS {
+	if ret != ERROR_INSUFFICIENT_SIZE {
```
klueska, line 1471: I had overloaded my suggested change in #11 without pointing this out explicitly:
```
	return Utilization[:ProcessSamplesCount], ret
```
But we should do that too.

elezar: OK, so the second call to `nvmlDeviceGetProcessUtilization` may also modify the `ProcessSamplesCount`? Or at the very least reduce it?

klueska: At least according to the API, yes.

klueska (approved):

gen/nvml/device.go
elezar: The docs:
```
 * On successful return, the function updates \a processSamplesCount with the number of process utilization sample
 * structures that were actually written. This may differ from a previously read value as instances are created or
 * destroyed.
 ```
don't seem to state that this is guaranteed to be less than the size of the input array, in which case I assume that returning `Utilization[:ProcessSamplesCount]` may be dangerous? Is it overkill to use `Utilization[:min(ProcessSamplesCount, len(Utilization)]` (or a more idiomatic version of this)?

klueska: It should be implied because it is a C API. The whole point of the dance to first get the sample count and then re-call the function with an appropriately allocated array is so that you have the chance to allocate an array that is big enough to hold all the samples.

PR #13, opened 2021-03-01
Regenerate bindings for DeviceGetProcessUtilization change

elezar:
This adds the regenerated bindings from #12 

Signed-off-by: Evan Lezar <elezar@nvidia.com>


pkg/nvml/nvml.go
```diff
@@ -25,2283 +25,2522 @@ package nvml
...
 
 // nvmlInit_v2 function as declared in nvml/nvml.h:1766
-func nvmlInit_v2() Return {
-	__ret := C.nvmlInit_v2()
-	__v := (Return)(__ret)
-	return __v
+func nvmlInit_v2() Return{
+__ret := C.nvmlInit_v2()
+__v := (Return)(__ret)
+return __v
```
klueska, line 33: What are all these changes in this file from?

pkg/nvml/types_gen.go
```diff
@@ -1,345 +0,0 @@
-// Code generated by cmd/cgo -godefs; DO NOT EDIT.
```
klueska, line 1: This file should definitely not be deleted....

elezar: Seems my docker vm has a full disk.

pkg/nvml/nvml.go
elezar: Was also due to the insufficient disk space. I missed the warnings / errors.

klueska (approved):

PR #14, opened 2021-03-01
Strip autogen timestamp

elezar:
This change ensures that the "timestamped" comment such as:
```
// WARNING: This file has automatically been generated on Tue, 16 Feb 2021 14:53:33 UTC.
```
in generated files is replaced with:
```
// WARNING: THIS FILE WAS AUTOMATICALLY GENERATED.
```
ensuring that unchanged files do not contribute to the diff.

The file modification timestamp should be sufficient to check when the file was modified.


Makefile
```diff
@@ -42,7 +42,9 @@ patch-nvml-h: $(PKG_BINDINGS_DIR)/nvml.h
 $(PKG_BINDINGS_DIR)/nvml.h: $(GEN_BINDINGS_DIR)/nvml.h | $(PKG_BINDINGS_DIR)
 	sed -E 's#(typedef\s+struct)\s+(nvml.*_st\*)\s+(nvml.*_t);#\1\n{\n    struct \2 handle;\n} \3;#g' $(<) > $(@)
 
-bindings: $(PKG_BINDINGS_DIR)/nvml.h $(SOURCES) | $(PKG_BINDINGS_DIR)
+bindings: .create-bindings .strip-autogen-comment
```
klueska, line 45: It probably doesn't matter, but will this break if we ever use `make -j` since `.create-bindings .strip-autogen-comment` will run in parallel?

Makefile
```diff
@@ -42,7 +42,9 @@ patch-nvml-h: $(PKG_BINDINGS_DIR)/nvml.h
...
+
+.create-bindings: $(PKG_BINDINGS_DIR)/nvml.h $(SOURCES) | $(PKG_BINDINGS_DIR)
```
klueska, line 47: What does the `|` do exactly? I've never used that before myself.

Makefile
```diff
@@ -51,6 +53,12 @@ bindings: $(PKG_BINDINGS_DIR)/nvml.h $(SOURCES) | $(PKG_BINDINGS_DIR)
 	cd -> /dev/null
 	rm -rf $(PKG_BINDINGS_DIR)/types.go $(PKG_BINDINGS_DIR)/_obj
 
+SED_SEARCH_STRING = WARNING: This file has automatically been generated on
+SED_REPLACE_STRING = WARNING: THIS FILE WAS AUTOMATICALLY GENERATED.
```
klueska, line 57: Does it make sense to use `:=` here since we don't need these vars to be regenerated based on future variable definitions.

Makefile
elezar: It shouldn't since `.strip-autogetn-comments` has an ordering dependency:
```
.strip-autogen-comment: | .create-bindings
```

Makefile
elezar: It's an ordering dependency. I suppose in this case it's not strictly needed and a normal dependency would work.

https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

Makefile
elezar: Sure. Will update as part of the rebase.

klueska (approved):

PR #15, opened 2021-03-01
Update device.GetUUID() to use v2 of the UUID buffer size

klueska:


elezar (approved):
Looks good. I like the lack of timestamp changes in the change set!

PR #20, opened 2021-06-02
Update NVML bindings to 11.2

elezar:
This change updates the NVML bindings to those from CUDA 11.2.0.

### DIFF
--- a/gen/nvml/init.go
+++ b/gen/nvml/init.go
@@ -90,6 +90,7 @@ var nvmlDeviceRemoveGpu = nvmlDeviceRemoveGpu_v1
 var nvmlDeviceGetGridLicensableFeatures = nvmlDeviceGetGridLicensableFeatures_v1
 var nvmlEventSetWait = nvmlEventSetWait_v1
 var nvmlDeviceGetAttributes = nvmlDeviceGetAttributes_v1
+var nvmlComputeInstanceGetInfo = nvmlComputeInstanceGetInfo_v1
 var nvmlDeviceGetComputeRunningProcesses = nvmlDeviceGetComputeRunningProcesses_v1
 var nvmlDeviceGetGraphicsRunningProcesses = nvmlDeviceGetGraphicsRunningProcesses_v1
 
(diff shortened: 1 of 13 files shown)


gen/nvml/init.go
```diff
@@ -145,6 +146,10 @@ func updateVersionedSymbols() {
 	if err == nil {
 		nvmlDeviceGetAttributes = nvmlDeviceGetAttributes_v2
 	}
+	err = nvml.Lookup("nvmlComputeInstanceGetInfo_v2")
+	if err == nil {
+		nvmlComputeInstanceGetInfo = nvmlComputeInstanceGetInfo_v2
+	}
```
klueska, line 152: There are some pieces missing to make this work correctly. For example, I don't see the old function called `nvmlComputeInstanceGetInfo()` renamed to `nvmlComputeInstanceGetInfo_v1()` anywhere, and I don't see a new implementation for `nvmlComputeInstanceGetInfo_v2()`.

elezar: Ah. That's a good point. Let me update.

elezar: Actually, no, I don't quite follow. Looking at other versioned APIs (e.g. `nvmlEventSetWait`, `nvmlDeviceGetGridLicensableFeatures`) there are not multiple implementations for different versions as long as the API stays the same (differences in returned structures not withstanding). In the case of `nvmlComputeInstanceGetInfo` we do set `_v1` as the default [here](https://github.com/NVIDIA/go-nvml/pull/20/files#diff-72259b77d95affcaab997579c53827abf1d307289d006c9b7576e5f2ecfe7af1R93) and also apply the c-for-go replacement rule [here](https://github.com/NVIDIA/go-nvml/pull/20/files#diff-a0750eafb313900829097ae6335f45f1c3b3dcfd706cf2214d7a98332ceb6a55R80).

klueska: Yeah, you are right. I am not sure what I was thinking of.  LGTM

PR #32, opened 2021-12-29
Fix a minor memory leak in `pkg/dl`

chenhengqi:


chenhengqi:
@klueska  @elezar  Please take a look.

elezar:
Thanks for the change @chenhengqi. This is a good catch. Would it be possible to split this PR into one that addresses the memory leak and one that introduces the idiomatic names for the constants (or at least have these as separate commits)?

chenhengqi:
I've just updated the changes to separate commit.

If this PR get merged to master, I hope you can also release a new tag so that I can vendor it locally. Thanks.

pkg/dl/dl.go
```diff
@@ -24,30 +24,43 @@ import (
...
 import "C"
 
+// RTLD constants
 const (
-	RTLD_LAZY = C.RTLD_LAZY
-	RTLD_NOW = C.RTLD_NOW
-	RTLD_GLOBAL = C.RTLD_GLOBAL
-	RTLD_LOCAL = C.RTLD_LOCAL
+	// Deprecated, don't use ALL_CAPS in Go names
+	RTLD_LAZY     = C.RTLD_LAZY
```
elezar, line 30: If we change the following line in `gen/init.go`:
```
	nvmlLibraryLoadFlags = dl.RTLD_LAZY | dl.RTLD_GLOBAL
```
We can remove the deprecated variable names here.

Or alternatively, what about rather adding a constant:
```
DefaultRtldFlags = C.RTLD_LAZY | C.RTLD_GLOBAL
```
here and using this instead?

gen/nvml/init.go
```diff
@@ -33,12 +33,14 @@ var nvml *dl.DynamicLibrary
 func Init() Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
 	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
+		log.Printf("error instantiating DynamicLibrary for %s\n", nvmlLibraryName)
```
elezar, line 36: Question: What is the benefit of using `log.Printf` over `fmt.Printf`?

gen/nvml/init.go
```diff
@@ -24,7 +24,7 @@ import "C"
 
 const (
 	nvmlLibraryName      = "libnvidia-ml.so.1"
-	nvmlLibraryLoadFlags = dl.RTLD_LAZY | dl.RTLD_GLOBAL
+	nvmlLibraryLoadFlags = dl.RtldLazy | dl.RtldGlobal
```
elezar, line 27: Update: OK, I see that the change was made here. See my earlier comment on adding something like `dl.DefaultRtldFlags` here.

gen/nvml/init.go
```diff
@@ -33,12 +33,14 @@ var nvml *dl.DynamicLibrary
 func Init() Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
 	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
+		log.Printf("error instantiating DynamicLibrary for %s\n", nvmlLibraryName)
+		return ERROR_UNKNOWN
```
elezar, line 37: I don't know whether I agree with returning `ERROR_UNKNOWN` here. This mixes errors that come from outside of NVML with NVML. What is the benefit of this over, for example, a caller recovering from the panic and printing the error there instead? We could provide a wrapper that does this if required?

```
// AssertInit calls the NVML function and recovers from any panics. An error is returned if
// a panic is detected or if the return from NVML.Init is not SUCCESS.
fun AssertInit() (err error) {
    
}
```

I would like to wait for @klueska to weigh in on this though.

elezar (commented):
Thanks for rearranging the commits!

pkg/dl/dl.go
chenhengqi: No sure whether you care about backward compatibility or not, so I just keep as it is.

gen/nvml/init.go
chenhengqi: It would add a datetime prefix.

gen/nvml/init.go
klueska: The reason a panic was used here is for exactly this reason. I didn't want to use an NVML error type to denote an error in loading the library.

On a related note, there have been several requests to allow this package to be built with compile-time linkage to `libnvidia-ml.so` instead of via `dllopen()`. I will need to think about the best way to support both options and the code here will likely change as a result.

gen/nvml/init.go
klueska: I'm OK using log over fmt but we should do it everywhere then (i.e. also in the `dl` library).

pkg/dl/dl.go
chenhengqi: Yes, you are right. The golang `debug/elf` package also contains such constants.

https://github.com/golang/go/blob/9de1ac6ac2cad3871760d0aa288f5ca713afd0a6/src/debug/elf/elf.go#L54-L62

chenhengqi:
@elezar @klueska 

Hello, I just keep the first two commits of the original PR.

Issue #31, opened 2021-11-29
General guidelines for integer types (`int` / `uint` / `uint32` / `uint64`) in manual wrappers


XuehaiPan:
Go is a strongly, statically typed language. It will not auto-cast values between integer types (`int` / `uint` / `uint32` / `uint64`) on return.

The bindings auto-generated by `c-for-go` always have a suffix `32` or `64` of `uint` types. But some manual wrappers cast the type to `int` or `uint` before returning while some others do not.

1. The binding receives `uint32`, the wrapper returns `int`:

    https://github.com/NVIDIA/go-nvml/blob/053ac6826f4cbf14b5197cc36a472278662a20b3/gen/nvml/device.go#L24-L29

2. The wrapper receives `int`, the binding receives `uint32`:

    https://github.com/NVIDIA/go-nvml/blob/053ac6826f4cbf14b5197cc36a472278662a20b3/gen/nvml/device.go#L31-L36

3. The binding and the wrapper both use `uint32`:

    https://github.com/NVIDIA/go-nvml/blob/053ac6826f4cbf14b5197cc36a472278662a20b3/gen/nvml/device.go#L383-L388

4. **Bug:** the binding requires `[]uint32` but `[]uint` is given:

    https://github.com/NVIDIA/go-nvml/blob/053ac6826f4cbf14b5197cc36a472278662a20b3/gen/nvml/device.go#L103-L109

klueska:
The goal was to have it be `int` when the exact type doesn't matter (to make it a bit more usable within existing Go programs that use `int` everyhwere). And to be specific to `uint32` and `uint64` when it does matter. We should never be using `uint`.

klueska:
Your (4.) is clearly a bug and should be `uint32`.

XuehaiPan:
> the reason it is being passed as a uint[] is because the underlying type is an unsigned long* which maps to uint[] in Go (which you can also see in the generated definition for this function)

I regenerated the bindings with the latest version of `c-for-go`, It gives me 3 bugs like this and the tests will not pass.

elezar:
Yes, we noted that an update to `c-for-go` would generate different types. We pin to the version [specified](https://github.com/NVIDIA/go-nvml/blob/master/Makefile#L17) as we didn't have capacity to properly investigate these at the time.

klueska:
[...]

PR #35, opened 2022-01-26
Fix argument type mismatching

jshen28:
Looks like there are few types mismatching. This
patch tries to correct them

Signed-off-by: shenjiatong <yshxxsjt715@gmail.com>


klueska:
There are bugs in the newer c-for-go that cause these type mis.matches (which is why we pin our version of c-for-go in the docker builds). Please see the following for reference:
https://github.com/NVIDIA/go-nvml/issues/31#issuecomment-981541876

PR #37, opened 2022-03-10
Bump to NVML version 11.3

klueska:


gen/nvml/device.go
```diff
@@ -1256,6 +1256,24 @@ func (Device Device) ResetGpuLockedClocks() Return {
 	return DeviceResetGpuLockedClocks(Device)
 }
 
+// nvmlDeviceSetMemoryLockedClocks()
+func DeviceSetMemoryLockedClocks(Device Device, MinMemClockMHz uint32, MaxMemClockMHz uint32) Return {
```
elezar, line 1260: I have probably asked this before, but are we explicitly using `uint32` here for `unsigned int` in C?

gen/nvml/init.go
```diff
@@ -93,6 +93,10 @@ var nvmlDeviceGetAttributes = nvmlDeviceGetAttributes_v1
 var nvmlComputeInstanceGetInfo = nvmlComputeInstanceGetInfo_v1
 var nvmlDeviceGetComputeRunningProcesses = nvmlDeviceGetComputeRunningProcesses_v1
 var nvmlDeviceGetGraphicsRunningProcesses = nvmlDeviceGetGraphicsRunningProcesses_v1
+var GetBlacklistDeviceCount = GetExcludedDeviceCount
```
elezar, line 96: Just to be clear, this is just a rename to allow this to be backward compatible?

gen/nvml/nvml.h
```diff
@@ -1,3 +1,5 @@
+/*** NVML VERSION: 11.3.58 ***/
```
elezar, line 1: Just a note. A really like this addition.

elezar (approved):
LGTM

gen/nvml/device.go
klueska: Only when it makes sense for the type. See the discussion here:
https://github.com/NVIDIA/go-nvml/issues/31

PR #38, opened 2022-03-10
Bump to NVML version 11.4

klueska:


gen/nvml/nvml.h
```diff
@@ -227,6 +229,18 @@ typedef struct nvmlBAR1Memory_st
     unsigned long long bar1Used;     //!< Allocated Used Memory (in bytes)
 }nvmlBAR1Memory_t;
 
+/**
+ * Information about running compute processes on the GPU, legacy version
+ * for older versions of the API.
+ */
+typedef struct nvmlProcessInfo_v1_st
```
elezar, line 236: This is what make the other commits redundant, correct?

elezar (commented):
Thanks. Glad to see the `v1` data structs.

elezar (approved):

gen/nvml/nvml.h
klueska: Yes. And I changed things around a bit to add a level of indirection through function pointers rather than branching in the wrapper function. This makes things cleaner as we get to 11.6 and there is yet another `ProcessInfo` version added.

PR #40, opened 2022-03-11
Bump to NVML version 11.6

klueska:


gen/nvml/vgpu.go
```diff
@@ -334,6 +334,26 @@ func (VgpuInstance VgpuInstance) GetGpuInstanceId() (int, Return) {
...
 }
 
+// nvml.VgpuInstanceGetGpuPciId()
+func VgpuInstanceGetGpuPciId(VgpuInstance VgpuInstance) (string, Return) {
+	var Length uint32 = 1 // Will be reduced upon returning
+	for {
+		VgpuPciId := make([]byte, Length)
+		ret := nvmlVgpuInstanceGetGpuPciId(VgpuInstance, &VgpuPciId[0], &Length)
+		if ret == SUCCESS {
+			return string(VgpuPciId[:clen(VgpuPciId)]), ret
```
elezar, line 344: Question, why use `clen` here instead of the returned value of `Length`? Or is that only set if it is too short? (the comments seem to indicate that this is the case)

gen/nvml/init.go
```diff
@@ -115,6 +117,20 @@ func (pis ProcessInfo_v1Slice) ToProcessInfoSlice() []ProcessInfo {
...
 }
 
+func (pis ProcessInfo_v2Slice) ToProcessInfoSlice() []ProcessInfo {
+	var newInfos []ProcessInfo
+	for _, pi := range pis {
+		info := ProcessInfo{
+			Pid:               pi.Pid,
+			UsedGpuMemory:     pi.UsedGpuMemory,
+			GpuInstanceId:     pi.GpuInstanceId,
+			ComputeInstanceId: pi.ComputeInstanceId,
```
elezar, line 127: Looking at this, the two types (`_v2` and unversioned) are the same struct as of this change?

gen/nvml/nvml.yml
```diff
@@ -82,4 +82,5 @@ TRANSLATOR:
       - {action: replace, from: "^nvmlDeviceGetGraphicsRunningProcesses$", to: "nvmlDeviceGetGraphicsRunningProcesses_v1"}
```
elezar, line 82: What about the `_v2` and `_v3` APIs now?

gen/nvml/device.go
```diff
@@ -2073,6 +2101,23 @@ func (GpuInstance GpuInstance) GetComputeInstanceProfileInfo(Profile int, EngPro
...
 
+// nvml.GpuInstanceGetComputeInstanceProfileInfoV()
+func GpuInstanceGetComputeInstanceProfileInfoV(GpuInstance GpuInstance, Profile int, EngProfile int, InfoV ComputeInstanceProfileInfoV) Return {
+	var Info *ComputeInstanceProfileInfo_v2
+	switch info := InfoV.(type) {
+	case *ComputeInstanceProfileInfo_v2:
+		info.Version = StructVersion.ComputeInstanceProfileInfo_v2()
+		Info = (*ComputeInstanceProfileInfo_v2)(unsafe.Pointer(info))
+	default:
+		return ERROR_INVALID_ARGUMENT
```
elezar, line 2112: Question: This is not an NVML error, should we include it here?

pkg/nvml/device.go
```diff
@@ -1848,6 +1970,23 @@ func (Device Device) GetGpuInstanceProfileInfo(Profile int) (GpuInstanceProfileI
 	return DeviceGetGpuInstanceProfileInfo(Device, Profile)
 }
 
+// nvml.DeviceGetGpuInstanceProfileInfoV()
+func DeviceGetGpuInstanceProfileInfoV(Device Device, Profile int, InfoV GpuInstanceProfileInfoV) Return {
```
elezar, line 1974: Question: Why can we not return the `GpuInstanceProfile` value here instead? (Like we do for the memory query).

gen/nvml/init.go
klueska: Yeah, I don't understand what they are trying to accomplish here. Maybe it was a mistake and the "v3" struct was supposed to have more fields. The comment seems to indicate so.

gen/nvml/vgpu.go
klueska: I think it is only set if it is too short (maybe). I know I ran into issues with other similar functions which is why I introduced the custom `clen` function to begin with. This follows the pattern of all other such functions.

gen/nvml/nvml.yml
klueska: What do you mean? In this file we just make sure that the "unversioned" symbol is converted to `v1` so that we can set the function pointer correctly in `init.go`. It follows the same pattern as all the other such functions.

gen/nvml/device.go
klueska: It basically is though. To me it feels equivalent to them setting an invalid `version` in the struct they pass in.

gen/nvml/nvml.yml
elezar: Ah you're right of course. Resolving.

pkg/nvml/device.go
klueska: If you look in nvml.h, in contrast to the Memory API (which has been versioned with a specific `_v2` suffix) this API is designed to be "future proof" with a generic suffix of `V`. That means that its basic signature won't change and apps compiled against older versions of NVML will continue to function properly on future versions of NVML. This will work because v2 (and all future versions) of the `GpuInstanceProfileInfo` struct will have a `version` field as its first field, and the API requires the caller to set this `version` before passing it to the function. This allows the underlying implementation to know which version of the struct is being passed in (e.g. a `v2` struct if compiled against an older version of NVML or a `v3` struct if compiled against a newer one).

This design is fine in the C world, but it causes problems in the Go world if we want some way for the user to (1) express their intent of which struct version they want, and (2) return structs of different types based on that intent. The best we can do is have the user pass a "number" for the desired struct version and then return an interface. But then it is up to the user to extract the dynamic type of this interface before they can use it....

The compromise I came to was to make the API look similar to the C API and have the struct populated after being passed as an input rather than a return value. If we are going to be in the business of having the user pass something to indicate what version of the struct they want to use, they might as well just pass the struct itself instead of passing a number and then having to extract the dynamic type from the value returned.

An alternative I considered was to just not bother mimicking this "generic" `V` style interface and instead just go with a `_v2` style API (like was done for memory), resulting in `DeviceGetGpuInstanceProfileInfo_v2()` as the function name. This would avoid this issue because the "intent" would be encoded in the function name instead of as a parameter and we would know exactly which type to return. I was worried this might clash with a name they introduce in the future though (since that is now versioning the API, not the struct -- and they may wish to version them separately if they, for example, add a parameter to the function at some point in the future).

I'm open to more ideas before we finalize this, but this was the "best" compromise I could come up with as I was playing with it.

klueska: What about doing this (this is tested and works):
```
// nvml.GpuInstanceGetComputeInstanceProfileInfoV()
type ComputeInstanceProfileInfoV struct{
    gpuInstance GpuInstance
    profile int
    engProfile int
}

func GpuInstanceGetComputeInstanceProfileInfoV(GpuInstance GpuInstance, Profile int, EngProfile int) ComputeInstanceProfileInfoV {
    return ComputeInstanceProfileInfoV{
        GpuInstance,
        Profile,
        EngProfile,
    }
}

func (GpuInstance GpuInstance) GetComputeInstanceProfileInfoV(Profile int, EngProfile int) ComputeInstanceProfileInfoV {
    return GpuInstanceGetComputeInstanceProfileInfoV(GpuInstance, Profile, EngProfile)
}

func (InfoV ComputeInstanceProfileInfoV) V2() (ComputeInstanceProfileInfo_v2, Return) {
    var Info ComputeInstanceProfileInfo_v2
    Info.SetVersion()
    ret := nvmlGpuInstanceGetComputeInstanceProfileInfoV(InfoV.gpuInstance, uint32(InfoV.profile), uint32(InfoV.engProfile), &Info)
    return Info, ret
}
```

With the previous iteration the caller would need to do the following:
```
var Info ComputeInstanceProfileInfo_v2
ret := GpuInstanceGetComputeInstanceProfileInfoV(GpuInstance, Profile, EngProfile, &Info)
if ret != nvml.SUCCESS {
    ...
}
```

With this new proposed way, this would turn into:
```
Info, ret := GpuInstanceGetComputeInstanceProfileInfoV(GpuInstance, Profile, EngProfile.V2()
if ret != nvml.SUCCESS {
    ...
}
```
Which is more similar to the way we typically do things (in terms of code flow), but the caller must be careful to make this extra `V2()` call at the end (or `V3()`, `V4()`, etc. in the future).

elezar: I like this more. The question is whether there is a way to have a "default" for this struct to avoid requiring that the user explicitly select a version?

gen/nvml/device.go
elezar: OK. That's fair enough.

pkg/nvml/device.go
klueska: Alternatively, we could break completely from having a `V` variant of this function and always just return the latest version (by attempting to get the latest -- and if that fails with an INVALID_ARGUMENT, go to the next previous version, and so on).

klueska: I don't know how to provide a `default` (or even implement my last suggestion) without returning an interface type though (because any such call would need to return one of multiple types depending on what the platform supports -- forcing the user to do a dynamic type assertion anyway).

klueska: I think I will go with what I have proposed here -- and even extend it to include `V1()` so that users can begin to "standardize" on using the `V` API for everything. I actually think it's fine that they pick a version and stick with it because the version they choose implies the minimum driver version their code will run on.

elezar (approved):

pkg/nvml/device.go
klueska: Updated

PR #43, opened 2022-04-20
Add support for adding libcuda bindings

elezar:
This change adds basic support for `libcuda.so` bindings. The only function that is currently supported is `cuDriverGetVersion`.

These bindings are useful on systems where NVML is not available and only basic query functionality is required.


klueska:
I'm not sure this is the right place for this. The hope was that this repo would eventually be taken over by the NVML team (which is why it is called `go-nvml` instead of something broader like `nvidia/go-bindings/pkg/nvml`, `nvidia/go-bindings/pkg/cuda`, etc.).

PR #47, opened 2022-07-13
Return error when library not found

LambdaHJ:
Signed-off-by: 黄金 <huangjin_henan@outlook.com>

Init function return an error when library open failed.

### DIFF
--- a/pkg/nvml/init.go
+++ b/pkg/nvml/init.go
@@ -32,13 +32,9 @@ var nvml *dl.DynamicLibrary
 // nvml.Init()
 func Init() Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
-	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
-	}
-
 	err := lib.Open()
 	if err != nil {
-		panic(fmt.Sprintf("error opening %s: %v", nvmlLibraryName, err))
+		return ERROR_LIBRARY_NOT_FOUND
 	}
 
 	nvml = lib
@@ -50,13 +46,9 @@ func Init() Return {
 // nvml.InitWithFlags()
 func InitWithFlags(Flags uint32) Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
-	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
-	}
-
 	err := lib.Open()
 	if err != nil {
-		panic(fmt.Sprintf("error opening %s: %v", nvmlLibraryName, err))
+		return ERROR_LIBRARY_NOT_FOUND
 	}
 
 	nvml = lib
--- a/gen/nvml/init.go
+++ b/gen/nvml/init.go
@@ -32,13 +32,9 @@ var nvml *dl.DynamicLibrary
 // nvml.Init()
 func Init() Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
-	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
-	}
-
 	err := lib.Open()
 	if err != nil {
-		panic(fmt.Sprintf("error opening %s: %v", nvmlLibraryName, err))
+		return ERROR_LIBRARY_NOT_FOUND
 	}
 
 	nvml = lib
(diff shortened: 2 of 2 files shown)


pkg/nvml/init.go
```diff
@@ -32,15 +35,10 @@ var nvml *dl.DynamicLibrary
...
 func Init() Return {
 	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
-	if lib == nil {
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
-	}
-
 	err := lib.Open()
 	if err != nil {
-		panic(fmt.Sprintf("error opening %s: %v", nvmlLibraryName, err))
+		return ERROR_LIBRARY_NOT_FOUND
```
elezar, line 40: Although this is cleaner than a `panic` from certain perspectives this is not a documented error returned by the underlying `nvmlInit` function.

What we tend to do internally is use a wrapper (e.g. `tryNvmlInit`) that recovers from the panic and converts the `Return` to an `error` instead.

With that said, I'll let @klueska weigh in on this change.

klueska: This comes up frequently enough as a perceived issue that I'm inclined to go ahead and change it since there is in fact an error called `ERROR_LIBRARY_NOT_FOUND`.

pkg/nvml/init.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/dl"
 )
 
+/*
+#include "nvml.h"
+*/
```
klueska, line 25: This is unnecessary, please remove.

elezar:
Thanks for the review @klueska.

@LambdaHJ please note that before we can merge this we would need to make the change in the correct place.

You would need to update the source in the `gen` folder [here](https://github.com/NVIDIA/go-nvml/blob/main/gen/nvml/init.go#L34), regenerate the bindings, and then also add the changes in `pkg` to Git.

To regenerate the bindings running:
```
make docker-bindings
```
should do this.

jasonliu747:
Hi @klueska @elezar , could you make a release after this PR? Really appreciated.

fmoessbauer:
> Hi @klueska @elezar , could you make a release after this PR? Really appreciated.

I just stumbled upon the same issue and was happy to see that there is already an MR. We also need this to test if NVML is available on the system.

klueska (approved):

klueska:
Looks good. @elezar can you do a quick test to make sure it works as expected before merging.

elezar:
Testing in a docker container on my mac (where I know there is no NVML).

On `main`:
```
$ make examples
$ ./devices
panic: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory

goroutine 1 [running]:
github.com/NVIDIA/go-nvml/pkg/nvml.Init(0xc000012000)
        /Users/elezar/src/go-nvml/pkg/nvml/init.go:41 +0x177
main.main()
        /Users/elezar/src/go-nvml/examples/devices/main.go:27 +0x48
```

With these changes:
```
$ ./devices
./devices: symbol lookup error: ./devices: undefined symbol: nvmlErrorString
```

Which is expected since the test code contains:
```
	ret := nvml.Init()
	if ret != nvml.SUCCESS {
		log.Fatalf("Unable to initialize NVML: %v", nvml.ErrorString(ret))
	}
```

Changing this to
```
	ret := nvml.Init()
	if ret == nvml.ERROR_LIBRARY_NOT_FOUND {
		log.Fatalf("Could not load NVML library")
	} else if ret != nvml.SUCCESS {
		log.Fatalf("Unable to initialize NVML: %v", nvml.ErrorString(ret))
	}
```
We see:
```
$ ./devices
2022/07/15 11:16:55 Could not load NVML library
```

@LambdaHJ one thing I noted is that running `make docker-bindings` shows:
```
diff --git a/pkg/nvml/init.go b/pkg/nvml/init.go
index 962823b..1572f81 100644
--- a/pkg/nvml/init.go
+++ b/pkg/nvml/init.go
@@ -36,6 +36,7 @@ func Init() Return {
        if err != nil {
                return ERROR_LIBRARY_NOT_FOUND
        }
+
        nvml = lib
        updateVersionedSymbols()

```
Could you run that again or add the newline as indicated?

pkg/nvml/init.go
```diff
@@ -32,15 +32,10 @@ var nvml *dl.DynamicLibrary
...
-		panic(fmt.Sprintf("error instantiating DynamicLibrary for %s", nvmlLibraryName))
-	}
-
 	err := lib.Open()
 	if err != nil {
-		panic(fmt.Sprintf("error opening %s: %v", nvmlLibraryName, err))
+		return ERROR_LIBRARY_NOT_FOUND
 	}
-
 	nvml = lib
```
elezar, line 39: Please add the newline above this.

elezar (changes requested):

pkg/nvml/init.go
LambdaHJ: done

elezar (approved):

PR #52, opened 2022-12-09
Bump to NVML version 11.7

klueska:


pkg/nvml/device.go
```diff
@@ -2286,3 +2286,145 @@ func DeviceGetBusType(Device Device) (BusType, Return) {
...
+	return int(PcieSpeed), ret
+}
+
+func (Device Device) GetPcieSpeed() (int, Return) {
+	return DeviceGetPcieSpeed(Device)
+}
+
+// nvml.DeviceGetGspFirmwareVersion()
+func DeviceGetGspFirmwareVersion(Device Device) (string, Return) {
+	Version := make([]byte, SYSTEM_DRIVER_VERSION_BUFFER_SIZE)
```
elezar, line 2345: SHould this be `NVML_GSP_FIRMWARE_VERSION_BUF_SIZE`?

klueska: Weird. I swear I looked for a const named like this one before choosing this alternative one and couldn't find it. Thanks for pointing it out.

klueska: Updated

klueska:
There is another issue in that `go tool cgo -godefs` does not handle anonymous nested structs and one was introduced in nvml.h for this version.

klueska:
Updated to handle addition of anonymous structs in nvml.h

Makefile
```diff
@@ -138,7 +143,7 @@ update-nvml-h:
 # Generate an image for containerized builds
 # Note: This image is local only
 .build-image: Dockerfile
-	$(DOCKER) build \
+	$(DOCKER) buildx build \
```
elezar, line 146: What motivated the switch to `buildx`?

elezar (approved):
Thanks @klueska. Looks good.

One minor question.

Makefile
klueska: `ARG	TARGETARCH`

Is not available with the normal build

klueska: I think it's fine to assume that buildx is available since this is only used to generate the bindings (and not something end users will ever need to run).

PR #53, opened 2022-12-10
Bump to NVML version 11.8

klueska:


gen/nvml/nvml.h
```diff
@@ -627,9 +627,9 @@ typedef struct
     struct
     {
         nvmlThermalController_t controller;
-        unsigned int defaultMinTemp;
-        unsigned int defaultMaxTemp;
-        unsigned int currentTemp;
+        int defaultMinTemp;
```
elezar, line 630: This is internal only from our perspective, correct?

elezar (commented):
Looks good, but let's wait for a rebase on #52 since I only looked at the last three commits.

gen/nvml/nvml.h
klueska: What do you mean by internal only? The file here in the `gen` directory should be a direct copy from what's upstream.

gen/nvml/gpm.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nvml
+
+// nvml.GpmMetricsGet()
```
elezar, line 17: Is this comment at the right place?

gen/nvml/gpm.go
```diff
@@ -0,0 +1,84 @@
...
+type GpmMetricsGetVType struct {
```
elezar, line 18: Where is this type actually used? Similarly for `GpmSupport*` below.

gen/nvml/nvml.h
elezar: I mean that the change of type doesn't require other changes in our wrapper.

gen/nvml/gpm.go
klueska: Yeah. I wanted it at the top of the full set of functions related to `GpmMetricsGet()`. I could move the actual `GpmMetricsGet()` to the top though if you think that's clearer.

gen/nvml/gpm.go
klueska: Yeah. I wanted it at the top of the full set of functions related to GpmMetricsGet(). I could move the actual GpmMetricsGet() to the top though if you think that's clearer.

elezar: I would think it's a little clearer in terms of docstring-like usage. It's not a blocker though.

gen/nvml/gpm.go
klueska: It's the return value of a call to `GpmMetricsGetV()`.

The usage is:
```
var MetricsGet GpmMetricsGetType)
ret := GpmMetricsGetV(&MetricsGet).V1()
```

elezar: I suppose my question is where do we expliclity need it, or is this something that a user needs to take care of. If that is the case we may want to add additional documentation to the affected functions.

klueska: It is the pattern I have been following for these new "versioned" structs that the NVML team has been introducing. The API using the latest struct version is always available using e.g. `GpmMetricsGet(...)`, but if you want to explicitly use a different version you can call e.g. `GpmMetricsGetV(...).V1()`, `GpmMetricsGetV(...).V2(), ... instead.

I first started doing it with NVMl 11.6 here:
https://github.com/NVIDIA/go-nvml/pull/40/files#diff-0298839f64113aa3439ac75706f86ef591ba5c852ef6a740e30fb52322471488R1974

gen/nvml/nvml.h
klueska: It shouldn't since they are struct fields set by the end user, not us.

gen/nvml/gpm.go
elezar: OK. So a user *could* use these once we have multiple versions but it is not required here.

elezar (approved):
I think my comments on docstrings are minor and should not block this.

PR #56, opened 2023-01-10
Bump to NVML version 12.0

klueska:


gen/nvml/device.go
```diff
@@ -2440,37 +2440,6 @@ func (Device Device) GetTargetFanSpeed(Fan int) (int, Return) {
 	return DeviceGetTargetFanSpeed(Device, Fan)
 }
 
-// nvml.DeviceGetPowerMode()
-func DeviceGetPowerMode(Device Device) (int, Return) {
```
elezar, line 2444: Just to clarify. These were removed.

elezar (approved):
Looks good.

gen/nvml/device.go
klueska: Yeah. I was surprised -- but I did a diff of the 11.8 and 12.0 nvml.h files and they are gone...

PR #69, opened 2023-07-23
Make dl package thread-safe

alexbagirov:
This PR is intended to make `dl` calls thread-safe. Problem might occur if, for example, `dlopen()` and `dlerror()` were run in different goroutines and there was an error which happened after `dlopen()` but was ignored.

To fix this behaviour, each dl call (`dlopen()`, `dlsym()`, `dlclose()`) runs in the same thread with its `dlerror()` call. This is done using `runtime.LockOSThread()` wrapper around those calls.

### DIFF
--- a/pkg/dl/dl.go
+++ b/pkg/dl/dl.go
@@ -104,6 +104,7 @@ func (dl *DynamicLibrary) Lookup(symbol string) error {
 
 	var pointer unsafe.Pointer
 	if err := withOSLock(func() error {
+		dlError()
 		pointer = C.dlsym(dl.handle, sym)
 		if pointer == nil {
 			return fmt.Errorf("symbol %q not found: %w", symbol, dlError())
--- a/pkg/dl/dl.go
+++ b/pkg/dl/dl.go
@@ -15,6 +15,7 @@
 package dl
 
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"unsafe"
@@ -49,19 +50,11 @@ func New(name string, flags int) *DynamicLibrary {
 	}
 }
 
-func useLibDL(action func() error) error {
+func withOSLock(action func() error) error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
 	return action()
-}
-
-type DLError struct {
-	err string
-}
-
-func (e DLError) Error() string {
-	return e.err
 }
 
 func dlError() error {
@@ -69,27 +62,23 @@ func dlError() error {
 	if lastErr == nil {
 		return nil
 	}
-	return DLError{
-		err: C.GoString(lastErr),
-	}
+	return errors.New(C.GoString(lastErr))
 }
 
 func (dl *DynamicLibrary) Open() error {
 	name := C.CString(dl.Name)
 	defer C.free(unsafe.Pointer(name))
 
-	var handle unsafe.Pointer
-	if err := useLibDL(func() error {
-		handle = C.dlopen(name, C.int(dl.Flags))
+	if err := withOSLock(func() error {
+		handle := C.dlopen(name, C.int(dl.Flags))
 		if handle == nil {
 			return dlError()
 		}
+		dl.handle = handle
 		return nil
 	}); err != nil {
 		return err
 	}
-
-	dl.handle = handle
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/dl/dl.go
```diff
@@ -48,35 +49,80 @@ func New(name string, flags int) *DynamicLibrary {
 	}
 }
 
+func useLibDL(action func() error) error {
```
elezar, line 52: I don't think `useLibDL` is a good name for this function. What about `withOSLock()` or `withSync()`?

pkg/dl/dl.go
```diff
@@ -48,35 +49,80 @@ func New(name string, flags int) *DynamicLibrary {
...
+		handle = C.dlopen(name, C.int(dl.Flags))
+		if handle == nil {
+			return dlError()
+		}
+		return nil
+	}); err != nil {
+		return err
 	}
+
 	dl.handle = handle
```
elezar, line 92: Is setting (and clearing) the handle not also something that should happen under the lock?

pkg/dl/dl.go
```diff
@@ -48,35 +49,80 @@ func New(name string, flags int) *DynamicLibrary {
...
 
 func (dl *DynamicLibrary) Lookup(symbol string) error {
 	sym := C.CString(symbol)
 	defer C.free(unsafe.Pointer(sym))
 
-	C.dlerror() // Clear out any previous errors
-	C.dlsym(dl.handle, sym)
-	err := C.dlerror()
-	if unsafe.Pointer(err) == C.NULL {
+	var pointer unsafe.Pointer
```
elezar, line 116: This pointer is local to the check below and could be initialised when calling `C.dlsym` below.

Also, we seem to have lost the call to `C.dlerror()` to clear previous errors. Is this no longer needed / expected?

elezar (changes requested):
Thanks for the PR. I have some minor comments / suggestions.

pkg/dl/dl.go
```diff
@@ -48,35 +49,80 @@ func New(name string, flags int) *DynamicLibrary {
...
+	return e.err
+}
+
+func dlError() error {
+	lastErr := C.dlerror()
+	if lastErr == nil {
+		return nil
+	}
+	return DLError{
+		err: C.GoString(lastErr),
```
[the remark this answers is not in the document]
alexbagirov: Done

pkg/dl/dl.go
```diff
@@ -48,35 +49,80 @@ func New(name string, flags int) *DynamicLibrary {
...
 }
 
+func useLibDL(action func() error) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	return action()
+}
+
+type DLError struct {
```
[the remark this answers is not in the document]
alexbagirov: Removed

pkg/dl/dl.go
alexbagirov: Renamed

pkg/dl/dl.go
alexbagirov: Moved under lock

pkg/dl/dl.go
alexbagirov: I see no cases where we have to clear `dlerror` since we call it each time error actually happens (and now we know that we won't miss any errors since they will arrive in the same thread as `open`/`close`/`sym` calls). So imo this is no longer needed

alexbagirov:
Fixed all the comments and also added one more test for `dl` module. Added `t.Parallel()` to each test since now we're sure they can be run simultaneously.

elezar (dismissed):
Thanks @alexbagirov.

This looks good now.

pkg/dl/dl.go
alexbagirov: Returned `dlError()` call.

pkg/dl/dl.go
```diff
@@ -72,11 +102,16 @@ func (dl *DynamicLibrary) Lookup(symbol string) error {
 	sym := C.CString(symbol)
 	defer C.free(unsafe.Pointer(sym))
 
-	C.dlerror() // Clear out any previous errors
-	C.dlsym(dl.handle, sym)
-	err := C.dlerror()
-	if unsafe.Pointer(err) == C.NULL {
+	var pointer unsafe.Pointer
+	if err := withOSLock(func() error {
+		dlError()
```
elezar, line 107: Could we keep the comment here indicating the reason why we are calling `dlError()` here?
```suggestion
		// Call dlError() to clear out any previous errors.
		dlError()
```

alexbagirov: Done

PR #81, opened 2023-10-12
Allow path to libnvidia-ml.so.1 to be specified

elezar:
These changes add a basic API for allowing the path to `libnvidia-ml.so.1` to be set.

This can be used in environments where the path to the library can be determined programatically but not reliably added to the `LD_LIBRARY_PATH` before program startup, for example.

This builds on the changes from #86

### DIFF
--- a/gen/nvml/lib.go
+++ b/gen/nvml/lib.go
@@ -27,33 +27,38 @@ import (
 import "C"
 
 const (
-	nvmlLibraryName      = "libnvidia-ml.so.1"
-	nvmlLibraryLoadFlags = dl.RTLD_LAZY | dl.RTLD_GLOBAL
+	defaultNvmlLibraryName      = "libnvidia-ml.so.1"
+	defaultNvmlLibraryLoadFlags = dl.RTLD_LAZY | dl.RTLD_GLOBAL
 )
 
 var errLibraryNotLoaded = errors.New("library not loaded")
+var errLibraryAlreadyLoaded = errors.New("library already loaded")
 
 // library represents an nvml library.
 // This includes a reference to the underlying DynamicLibrary
 type library struct {
 	sync.Mutex
-	dl dynamicLibrary
+	path  string
+	flags int
+	dl    dynamicLibrary
 }
 
 // libnvml is a global instance of the nvml library.
-var libnvml library
+var libnvml = library{
+	path:  defaultNvmlLibraryName,
+	flags: defaultNvmlLibraryLoadFlags,
+}
 
 var _ Interface = (*library)(nil)
 
-// Default returns a reference to the global library instance.
-// This is returned as an interface so as to not directly expose the underlying instance type.
-func Default() Interface {
-	return &libnvml
+// GetLibrary returns a the library as a Library interface.
+func (l *library) GetLibrary() Library {
+	return l
 }
 
-// Library returns a representation of the underlying library.
-func (l *library) Library() Library {
-	return l
+// GetLibrary returns a representation of the underlying library that implements the Library interface.
+func GetLibrary() Library {
+	return libnvml.GetLibrary()
 }
 
 // Lookup checks whether the specified library symbol exists in the library.
@@ -66,8 +71,8 @@ func (l *library) Lookup(name string) error {
 }
 
 // newDynamicLibrary is a function variable that can be overridden for testing.
-var newDynamicLibrary = func(name string, flags int) dynamicLibrary {
-	return dl.New(name, flags)
+var newDynamicLibrary = func(path string, flags int) dynamicLibrary {
+	return dl.New(path, flags)
 }
 
 // load initializes the library and updates the versioned symbols.
@@ -79,10 +84,10 @@ func (l *library) load() error {
 		return nil
 	}
 
-	dl := newDynamicLibrary(nvmlLibraryName, nvmlLibraryLoadFlags)
+	dl := newDynamicLibrary(l.path, l.flags)
 	err := dl.Open()
 	if err != nil {
-		return err
+		return fmt.Errorf("error opening %s: %w", l.path, err)
 	}
 
 	l.dl = dl
@@ -91,7 +96,9 @@ func (l *library) load() error {
 	return nil
 }
 
-// close the associated dynamic library if required.
+// close the underlying library and ensure that the global pointer to the
+// library is set to nil to ensure that subsequent calls to open will reinitialize it.
+// Multiple calls to an already closed nvml library will return without error.
 func (l *library) close() error {
 	l.Lock()
 	defer l.Unlock()
@@ -102,7 +109,7 @@ func (l *library) close() error {
 
 	err := l.dl.Close()
 	if err != nil {
-		return fmt.Errorf("error closing %s: %w", nvmlLibraryName, err)
+		return fmt.Errorf("error closing %s: %w", l.path, err)
 	}
 
 	l.dl = nil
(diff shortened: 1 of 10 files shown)


gen/nvml/lib.go
```diff
@@ -25,7 +30,64 @@ const (
 	nvmlLibraryLoadFlags = dl.RTLD_LAZY | dl.RTLD_GLOBAL
 )
 
-var nvml *dl.DynamicLibrary
+type nvmlLib struct {
+	mu  sync.Mutex
```
klueska, line 35: Is there a particular reason to include the `mu`here?

For types like these, I typically just embed the mutex type so I can do e.g. `nvmlLib.Lock()` instead of `nvmlLib.mu.Lock()`

gen/nvml/lib.go
```diff
@@ -25,7 +30,64 @@ const (
...
+	lib *dl.DynamicLibrary
+}
+
+// nvml represents the global nvml library.
+var nvml nvmlLib
+
+// Open ensures that the nvml library is initialized.
+// If it is not initialized, the library is opened using dlopen and the versioned
+// symbols are updated.
+func (l *nvmlLib) Open() error {
```
klueska, line 50: Given this comment, should be be called `init` instead of `open`? Or should the comment be updated? 

Probably just the comment updated to:
```
// Open ensures that the nvml library is open and initialized with the proper versioned symbols.
// Multiple calls to an already opened nvml library will return without error.
```

gen/nvml/lib.go
```diff
@@ -25,7 +30,64 @@ const (
...
+
+	l.updateVersionedSymbols()
+
+	return nil
+}
+
+// Close uninitialized the nvml library if required.
+// This closes the underlying library and ensures that the global pointer to the
+// library is set to nil to ensure that subsequent calls to init will reinitialize it.
+func (l *nvmlLib) Close() error {
```
klueska, line 73: likewise here ... maybe just update the comment to:
```
// Close closes the underlying library and ensures that the global pointer to the
// library is set to nil to ensure that subsequent calls to open will reinitialize it.
// Multiple calls to an already closed nvml library will return without error.
```

pkg/dl/dl.go
```diff
@@ -36,16 +36,22 @@ const (
 	RTLD_DEEPBIND = C.RTLD_DEEPBIND
 )
 
+type Config struct {
+	Name  string
+	Flags int
+}
+
```
klueska, line 43: Just curious, why was this level of indirection added? Having to call this feels awkward to me, for example: `nvml.Config.Name`.

I'd almost rather just see the global variable we have renamed to `nvmllib` and then leave this as is, so we get `nvmllib.Name`.

That said -- maybe `Name` isn't the right thing anymore given what we are trying to accomplish in this PR -- and `Path`  is probably better.

gen/nvml/lib.go
```diff
@@ -234,3 +245,29 @@ func (l *nvmlLib) updateVersionedSymbols() {
 		nvmlVgpuInstanceGetLicenseInfo = nvmlVgpuInstanceGetLicenseInfo_v2
 	}
 }
+
+// Option represents a functional option to configure the underlying NVML library
+type Option func(*nvmlLib)
+
+// WithLibName provides an option to set the library name to be used by the NVML library.
+func WithLibName(libName string) Option {
```
klueska, line 253: `WithLibraryPath()`?

gen/nvml/lib.go
```diff
@@ -234,3 +245,29 @@ func (l *nvmlLib) updateVersionedSymbols() {
...
+	return func(l *nvmlLib) {
+		l.Config.Name = libName
+	}
+}
+
+// SetLibOptions applies the specified options to the NVML library.
+func SetLibOptions(opts ...Option) {
```
klueska, line 260: `SetLibraryOptions()`?

gen/nvml/lib.go
```diff
@@ -234,3 +245,29 @@ func (l *nvmlLib) updateVersionedSymbols() {
 		nvmlVgpuInstanceGetLicenseInfo = nvmlVgpuInstanceGetLicenseInfo_v2
 	}
 }
+
+// Option represents a functional option to configure the underlying NVML library
+type Option func(*nvmlLib)
```
klueska, line 250: `LibraryOption`?

gen/nvml/lib.go
elezar: Using `mu` is pretty common in example code. Maybe that's for demonstration purposes though.

I will update to embed the type then.

gen/nvml/lib.go
elezar: Yes, I think `init` or `initLibrary` is better here. Note that I have created #86 which already includes this (or something similar) for the `Lookup` changes we discussed.

gen/nvml/lib.go
elezar: Should be addressed in the latest.

gen/nvml/lib.go
elezar: I elected to go with `library.load()` here. See the changes in #86.

pkg/dl/dl.go
elezar: Also, I have not yet changed the `dl` package. I will rather do that in a follow-up.

gen/nvml/lib.go
elezar: Updated.

pkg/nvml/lib.go
```diff
@@ -255,3 +268,36 @@ func (l *library) updateVersionedSymbols() {
...
+	}
+
+	for _, opt := range opts {
+		opt(&libnvml)
+	}
+
+	if libnvml.path == "" {
+		libnvml.path = defaultNvmlLibraryName
+	}
+	if libnvml.flags == 0 {
```
klueska, line 298: How do you unset all flags?

gen/nvml/lib.go
```diff
@@ -66,8 +72,8 @@ func (l *library) Lookup(name string) error {
 }
 
 // newDynamicLibrary is a function variable that can be overridden for testing.
-var newDynamicLibrary = func(name string, flags int) dynamicLibrary {
-	return dl.New(name, flags)
+var newDynamicLibrary = func(path string, flags int) dynamicLibrary {
+	return dl.New(path, flags)
```
klueska, line 76: I guess the `dl` library is already capable of handling a full path, we just never used this until now?

gen/nvml/lib.go
```diff
@@ -255,3 +268,36 @@ func (l *library) updateVersionedSymbols() {
...
+// WithLibraryPath provides an option to set the library name to be used by the NVML library.
+func WithLibraryPath(path string) LibraryOption {
+	return func(l *library) {
+		l.path = path
+	}
+}
+
+// SetLibraryOptions applies the specified options to the NVML library.
+// If this is called when a library is already loaded, and error is raised.
+func SetLibraryOptions(opts ...LibraryOption) error {
```
klueska, line 284: I think it was in the last MR, but can you explain the purpose of the `Default()` method? I'm just thinking about it from a callsite perspective --- `nvml.Default()` doesn't really imply to me that I'm getting a reference to the default implementation of the nvml Interface. When / how do you actually expect this to be used?

pkg/nvml/lib.go
elezar: Is that a use case that is expected? If so we should use a `*int` to detect whether it is set or not.

Note it's not currently possible to set the flags through the options, so we could address this in a follow-up.

gen/nvml/lib.go
elezar: Yes. When `dl.Open` is called, the `path` is forwarded to `C.dlopen` which has well defined semantics for handling the path and where to search.

pkg/nvml/lib.go
klueska: Note sure. I doubt anyone will ever override this (beyond adding to it) so its probably fine.

gen/nvml/lib.go
klueska: I see -- its specifically so we can get at that Library method -- can we just make a top level Library() function then?

elezar: Would `GetLibrary()` be ok? We would either need to rename this or the `Library` interface.

klueska: Hmm, right -- I'd rather the top level call be the same as what we would eventually call on an instance of `Interface`. Maybe that becomes `GetLibrary()` as well?

elezar: OK. I renamed the function `GetLibrary` and also included an implementation at the package level.

klueska (approved):

pkg/nvml/lib.go
elezar: Let's reevaluate if it ever becomes relevant.

PR #84, opened 2023-10-13
Add top-level function to lookup symbols

elezar:
This change adds a top-level LookupSymbol function. This replaces the need to explicitly open a library locally and check the symbols there in clients of these bindings.

Note that this API is considered experimental.

### DIFF
--- a/gen/nvml/lib.go
+++ b/gen/nvml/lib.go
@@ -1,0 +1,30 @@
+/**
+# Copyright 2023 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


gen/nvml/lib.go
```diff
@@ -0,0 +1,30 @@
...
+// LookupSymbol provides top-level access to symbol lookup in the configured library.
+// Note that this requires that the library be loaded as is done in a call to nvml.Init().
+//
+// Deprecated: LookupSymbol represents an unstable API and should be considered experimental.
+func LookupSymbol(name string) error {
+	if nvml == nil {
+		return fmt.Errorf("error looking up %s: library %s not initialized", name, nvmlLibraryName)
+	}
+	return nvml.Lookup(name)
+}
```
klueska, line 30: Does it make more sense to provide a function that just returns the `nvml` library itsself and then allow them to call `Lookup()` on that (or whatever other operation we may have on the `DynamicLibrary` type in the future)?

PR #86, opened 2023-10-17
Rework Lookup interface

elezar:
This change reworks the Lookup function added in #84.

It adds an Interface type at a top level with `Lookup` functionality in a `Library` interface.


gen/nvml/api.go
```diff
@@ -0,0 +1,28 @@
...
+package nvml
+
+// Info represents a set of additional functions that are not associated with the
+// NVML library as pure bindings.
+type Info interface {
+	Lookup(string) error
+}
+
+// Interface represents the interace for the NVML library.
+type Interface interface {
```
elezar, line 26: @klueska as discussed offline, this is the type that would be extended to include our top-level NVML functions, with the `Info` interface capturing the non-NVML functions such as `Lookup` and any other functions that we require (read-only).

klueska: In general, I like this approach, but I'm not sure of the name `Info` -- at least not paired with the only existing function of `Lookup`. 

A typical usage would then look like:
```
struct mytype {
    nvml nvml.Interface
}
mytype.nvml = nvml.New(WithLibraryPath("..."))
mytype.nvml.Info().Lookup("someSymbol")
```
One of the two alternatives would be better in my view:
```
mytype.nvml.Library().LookupSymbol("someSymbol")
```
or
```
mytype.nvml.Info().LookupLibrarySymbol("someSymbol")
```

elezar: I think being less generic and using `Library()` over `Info()` is better. We could always define multiple interfaces which group functions if we need to extend this.

elezar:
@klueska  I've reworked this a bit. This now incorporates some of the changes from #81 and should make those changes cleaner too. I am addressing the test failures in #89 as a separate PR.

gen/nvml/api.go
```diff
@@ -0,0 +1,37 @@
...
+// Library defines a set of functions defined on the underlying dynamic library.
+type Library interface {
+	Lookup(string) error
+}
+
+// dynamicLibrary is an interface for abstacting the underlying library.
+// This also allows for mocking and testing.
+
+//go:generate moq -stub -out dynamicLibrary_mock.go . dynamicLibrary
+type dynamicLibrary interface {
```
klueska, line 28: Is this solely so we can mock it with moq?

gen/nvml/init.go
```diff
@@ -16,43 +16,23 @@ package nvml
...
-)
-
-var nvml *dl.DynamicLibrary
-
 // nvml.Init()
 func Init() Return {
-	lib := dl.New(nvmlLibraryName, nvmlLibraryLoadFlags)
-	err := lib.Open()
-	if err != nil {
+	if err := lib.load(); err != nil {
```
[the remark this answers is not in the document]
elezar: Yes, this is what was previously `nvml`. Any suggestions? `libnvidiaML`?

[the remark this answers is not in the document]
klueska: yeah, that seems reasonable -- or even just `libnvml`

klueska (dismissed):

klueska (approved):

PR #91, opened 2023-10-24
Add golangci lint

elezar:
This PR adds an initial golangci lint configuration and also addresses some minor lint errors.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,17 @@
+# please keep this alphabetized
+linters:
+  enable: []
+  disable:
+    - asciicheck
+    - contextcheck
+    - forcetypeassert
+    - gocritic
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    - stylecheck
+
+run:
+  tests: true
+  timeout: 10m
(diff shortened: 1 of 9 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - gocritic
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    - stylecheck
+
+run:
+  tests: true
+  timeout: 1m
```
ArangoGutierrez, line 17: is 1 min enough? maybe 5 or 10, github action instances are slow some times

ArangoGutierrez: oh I see everything is disabled, then 1 min is ok for now

ArangoGutierrez (approved):
/lgtm

.golangci.yaml
elezar: Note. There are still some linters that are enabled by default: `errcheck` and `gosimple` for example.

These are the ones that trigger the other fixes required in this changset.

I have bumped the timeout to 10m though.

ArangoGutierrez (approved):

.golangci.yaml
ArangoGutierrez: DO you think, the lint check should be an extra github action?

PR #99, opened 2024-02-12
Add fallback ErrorString function

### LINKED ISSUE #82: nvmlErrorString error
i import nvml in my code，but when run code，have this error
image

@klueska

elezar:
This change adds a fallback ErrorString function that is used when the nvmlErrorString function cannot be resolved. For example, when the library is not loaded.

Closes #82

### DIFF
--- a/gen/nvml/return.go
+++ b/gen/nvml/return.go
@@ -14,7 +14,80 @@
 
 package nvml
 
+import (
+	"fmt"
+)
+
 // nvml.ErrorString()
-func ErrorString(Result Return) string {
-	return nvmlErrorString(Result)
+func ErrorString(r Return) string {
+	if err := GetLibrary().Lookup("nvmlErrorString"); err != nil {
+		return nvmlErrorString(r)
+	}
+	return fallbackErrorStringFunc(r)
+}
+
+// fallbackErrorStringFunc provides a basic nvmlErrorString implementation.
+// This allows the nvml.ErrorString function to be used even if the NVML library
+// is not loaded.
+var fallbackErrorStringFunc = func(r Return) string {
+	switch r {
+	case SUCCESS:
+		return "SUCCESS"
+	case ERROR_UNINITIALIZED:
+		return "ERROR_UNINITIALIZED"
+	case ERROR_INVALID_ARGUMENT:
+		return "ERROR_INVALID_ARGUMENT"
+	case ERROR_NOT_SUPPORTED:
+		return "ERROR_NOT_SUPPORTED"
+	case ERROR_NO_PERMISSION:
+		return "ERROR_NO_PERMISSION"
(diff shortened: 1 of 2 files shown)


klueska:
Does this mimic what we did in go-nvlib?

elezar:
> Does this mimic what we did in go-nvlib?

The `fallbackErrorString` function is the same. The mechanics is different since I don't set a global function var but do a symbol lookup with every call. I thought that this was simpler since it didn't require state (the global) function pointer to be updated.

Note that when we originally implemented the logic in `go-nvlib`, we didn't have the `Lookup` function exposed at a package level here.

klueska:
Seems reasonable. Given that we only end up calling this on error conditions, the extra bit of time to look up the symbol should be negligible.

klueska (dismissed):

pkg/nvml/return.go
```diff
@@ -14,7 +14,80 @@
...
+	case ERROR_FREQ_NOT_SUPPORTED:
+		return "ERROR_FREQ_NOT_SUPPORTED"
+	case ERROR_ARGUMENT_VERSION_MISMATCH:
+		return "ERROR_ARGUMENT_VERSION_MISMATCH"
+	case ERROR_DEPRECATED:
+		return "ERROR_DEPRECATED"
+	case ERROR_UNKNOWN:
+		return "ERROR_UNKNOWN"
+	default:
+		return fmt.Sprintf("Unknown return value: %d", r)
```
klueska, line 91:
```suggestion
		return fmt.Sprintf("unknown return value: %d", r)
```

gen/nvml/return.go
```diff
@@ -14,7 +14,80 @@
...
+	case ERROR_FREQ_NOT_SUPPORTED:
+		return "ERROR_FREQ_NOT_SUPPORTED"
+	case ERROR_ARGUMENT_VERSION_MISMATCH:
+		return "ERROR_ARGUMENT_VERSION_MISMATCH"
+	case ERROR_DEPRECATED:
+		return "ERROR_DEPRECATED"
+	case ERROR_UNKNOWN:
+		return "ERROR_UNKNOWN"
+	default:
+		return fmt.Sprintf("Unknown return value: %d", r)
```
klueska, line 91:
```suggestion
		return fmt.Sprintf("unknown return value: %d", r)
```

klueska (approved):

yxxhero:
@elezar @klueska 
```
// nvml.ErrorString()
func ErrorString(r Return) string {
	if err := GetLibrary().Lookup("nvmlErrorString"); err == nil {
		return nvmlErrorString(r)
	}
	return fallbackErrorStringFunc(r)
}
```
I think that it should use `err == nil`.

elezar:
Yes, you're right. Thanks for your PR. I think it would be better to invert the return statements.

PR #105, opened 2024-03-14
Add a refcount to dl load() and close() calls

klueska:


gen/nvml/lib.go
```diff
@@ -77,10 +78,14 @@ var newDynamicLibrary = func(path string, flags int) dynamicLibrary {
 
 // load initializes the library and updates the versioned symbols.
 // Multiple calls to an already loaded library will return without error.
-func (l *library) load() error {
+func (l *library) load() (rerr error) {
 	l.Lock()
 	defer l.Unlock()
-	if l.dl != nil {
+
+	l.refcount.Inc()
```
elezar, line 85: Question: Does something like:
```
defer l.refcount.IncOnNoError(rerror)
```
make sense here. (adjusting the `l.refcount > 1` of course).

Also, are we assured that this defer would execute before the unlock?

gen/nvml/lib.go
```diff
@@ -99,10 +104,17 @@ func (l *library) load() error {
 // close the underlying library and ensure that the global pointer to the
 // library is set to nil to ensure that subsequent calls to open will reinitialize it.
 // Multiple calls to an already closed nvml library will return without error.
-func (l *library) close() error {
+func (l *library) close() (rerr error) {
 	l.Lock()
 	defer l.Unlock()
 
+	l.refcount.Dec()
+	defer l.refcount.IncOnError(rerr)
```
elezar, line 112: What about when `refcount == 0` to start with and then does not get decreased to `-1` (because we're not allowing negative values). If there is then an error, our post-condition for calling `close` is then that `refcount == 1` and not `refcount == 0`.

gen/nvml/refcount.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (r *refcount) Inc() {
+	(*r)++
+}
+
+func (r *refcount) Dec() {
+	if *r > 0 {
+		(*r)--
+	}
+}
```
elezar, line 31: What about:
```suggestion
func (r *refcount) Dec() func(error) {
	if *r > 0 {
		(*r)--
		return r.IncOnError
	}
	// return a no-op if we don't decrease refcount.
	return func(error) {}
}
```

And then we would use this:
```
defer l.refcount.Dec()(rerr)
```
?

elezar (commented):
I think this looks good. I thing our negative check and handling errors in this case need to be adjusted. 

Added a suggestion for how to deal with this -- although I understand if we think it's a bit convoluted.

gen/nvml/refcount.go
klueska: This is less readable to me than just having two separate calls.

elezar: I meant that this handles the zero case, but sure.

gen/nvml/lib_test.go
```diff
@@ -134,6 +134,43 @@ func TestLookupFromDefault(t *testing.T) {
 	}
 }
 
+func TestLoadAndCloseNesting(t *testing.T) {
+	var numOpens, numCloses int
+
+	dl := &dynamicLibraryMock{
+		OpenFunc: func() error {
+			numOpens++
```
elezar, line 142: I don't think this is required. You could use `len(dl.calls.Open)` and `len(dl.calls.Close)` directly.

gen/nvml/lib_test.go
```diff
@@ -134,6 +134,43 @@ func TestLookupFromDefault(t *testing.T) {
...
+		},
+		CloseFunc: func() error {
+			numCloses++
+			return nil
+		},
+	}
+
+	defer setNewDynamicLibraryDuringTest(dl)()
+	defer resetLibrary()
+	GetLibrary()
```
elezar, line 153: Is the `GetLibrary()` call required?

gen/nvml/lib_test.go
```diff
@@ -134,6 +134,43 @@ func TestLookupFromDefault(t *testing.T) {
...
+
+	// When calling load twice, the library was only opened once
```
elezar, line 155: Could we add the following initial check too:
```
	// When calling close before opening the library nothing happens.
	require.Equal(t, 0, len(dl.calls.Close))
	require.Nil(t, libnvml.close())
	require.Equal(t, 0, len(dl.calls.Close))
```
to ensure that we don't try to close a nil library?

pkg/nvml/lib_test.go
```diff
@@ -134,6 +134,43 @@ func TestLookupFromDefault(t *testing.T) {
 	}
 }
 
+func TestLoadAndCloseNesting(t *testing.T) {
+	var numOpens, numCloses int
+
+	dl := &dynamicLibraryMock{
```
elezar, line 140: We are missing test cases for when `dl.Open` or `dl.Close` returns errors.

elezar (commented):
Thanks for this @klueska

Minor comments on testing top-level failure cases and using the mock properties instead of tracking calls explicitly.

pkg/nvml/lib_test.go
elezar: I added 
```
diff --git a/pkg/nvml/lib_test.go b/pkg/nvml/lib_test.go
index a800b7e..95a7f6a 100644
--- a/pkg/nvml/lib_test.go
+++ b/pkg/nvml/lib_test.go
@@ -135,15 +135,11 @@ func TestLookupFromDefault(t *testing.T) {
 }
 
 func TestLoadAndCloseNesting(t *testing.T) {
-	var numOpens, numCloses int
-
 	dl := &dynamicLibraryMock{
 		OpenFunc: func() error {
-			numOpens++
 			return nil
 		},
 		CloseFunc: func() error {
-			numCloses++
 			return nil
 		},
 	}
@@ -152,23 +148,102 @@ func TestLoadAndCloseNesting(t *testing.T) {
 	defer resetLibrary()
 	GetLibrary()
 
+	// When calling close before opening the library nothing happens.
+	require.Equal(t, 0, len(dl.calls.Close))
+	require.Nil(t, libnvml.close())
+	require.Equal(t, 0, len(dl.calls.Close))
+
 	// When calling load twice, the library was only opened once
-	require.Equal(t, 0, numOpens)
+	require.Equal(t, 0, len(dl.calls.Open))
 	require.Nil(t, libnvml.load())
-	require.Equal(t, 1, numOpens)
+	require.Equal(t, 1, len(dl.calls.Open))
 	require.Nil(t, libnvml.load())
-	require.Equal(t, 1, numOpens)
+	require.Equal(t, 1, len(dl.calls.Open))
 
 	// Only after calling close twice, was the library closed
-	require.Equal(t, 0, numCloses)
+	require.Equal(t, 0, len(dl.calls.Close))
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 0, numCloses)
+	require.Equal(t, 0, len(dl.calls.Close))
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 1, numCloses)
+	require.Equal(t, 1, len(dl.calls.Close))
 
 	// Calling close again doesn't attempt to close the library again
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 1, numCloses)
+	require.Equal(t, 1, len(dl.calls.Close))
+}
+
+func TestLoadAndCloseWithErrors(t *testing.T) {
+	testCases := []struct {
+		description           string
+		dl                    dynamicLibrary
+		expectedLoadRefcount  refcount
+		expectedCloseRefcount refcount
+	}{
+		{
+			description: "regular flow",
+			dl: &dynamicLibraryMock{
+				OpenFunc: func() error {
+					return nil
+				},
+				CloseFunc: func() error {
+					return nil
+				},
+			},
+			expectedLoadRefcount:  1,
+			expectedCloseRefcount: 0,
+		},
+		{
+			description: "open error",
+			dl: &dynamicLibraryMock{
+				OpenFunc: func() error {
+					return errors.New("")
+				},
+				CloseFunc: func() error {
+					return nil
+				},
+			},
+			expectedLoadRefcount:  0,
+			expectedCloseRefcount: 0,
+		},
+		{
+			description: "close error",
+			dl: &dynamicLibraryMock{
+				OpenFunc: func() error {
+					return nil
+				},
+				CloseFunc: func() error {
+					return errors.New("")
+				},
+			},
+			expectedLoadRefcount:  1,
+			expectedCloseRefcount: 1,
+		},
+		{
+			description: "open and close error",
+			dl: &dynamicLibraryMock{
+				OpenFunc: func() error {
+					return errors.New("")
+				},
+				CloseFunc: func() error {
+					return errors.New("")
+				},
+			},
+			expectedLoadRefcount:  0,
+			expectedCloseRefcount: 0,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			defer setNewDynamicLibraryDuringTest(tc.dl)()
+			defer resetLibrary()
+
+			_ = libnvml.load()
+			require.Equal(t, tc.expectedLoadRefcount, libnvml.refcount)
+			_ = libnvml.close()
+			require.Equal(t, tc.expectedCloseRefcount, libnvml.refcount)
+		})
+	}
 }
 
 func setNewDynamicLibraryDuringTest(dl dynamicLibrary) func() {
```

gen/nvml/lib_test.go
elezar:
```
diff --git a/pkg/nvml/lib_test.go b/pkg/nvml/lib_test.go
index a800b7e..ad5a01e 100644
--- a/pkg/nvml/lib_test.go
+++ b/pkg/nvml/lib_test.go
@@ -135,15 +135,11 @@ func TestLookupFromDefault(t *testing.T) {
 }
 
 func TestLoadAndCloseNesting(t *testing.T) {
-	var numOpens, numCloses int
-
 	dl := &dynamicLibraryMock{
 		OpenFunc: func() error {
-			numOpens++
 			return nil
 		},
 		CloseFunc: func() error {
-			numCloses++
 			return nil
 		},
 	}
@@ -153,22 +149,22 @@ func TestLoadAndCloseNesting(t *testing.T) {
 	GetLibrary()
 
 	// When calling load twice, the library was only opened once
-	require.Equal(t, 0, numOpens)
+	require.Equal(t, 0, len(dl.calls.Open))
 	require.Nil(t, libnvml.load())
-	require.Equal(t, 1, numOpens)
+	require.Equal(t, 1, len(dl.calls.Open))
 	require.Nil(t, libnvml.load())
-	require.Equal(t, 1, numOpens)
+	require.Equal(t, 1, len(dl.calls.Open))
 
 	// Only after calling close twice, was the library closed
-	require.Equal(t, 0, numCloses)
+	require.Equal(t, 0, len(dl.calls.Close))
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 0, numCloses)
+	require.Equal(t, 0, len(dl.calls.Close))
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 1, numCloses)
+	require.Equal(t, 1, len(dl.calls.Close))
 
 	// Calling close again doesn't attempt to close the library again
 	require.Nil(t, libnvml.close())
-	require.Equal(t, 1, numCloses)
+	require.Equal(t, 1, len(dl.calls.Close))
 }
 
 func setNewDynamicLibraryDuringTest(dl dynamicLibrary) func() {
```

elezar (dismissed):
Just to clarify, I think the additions that I proposed are fine as follow-ups.

gen/nvml/lib_test.go
klueska: Integrating ...

pkg/nvml/lib_test.go
klueska: Integrating...

elezar (approved):

Issue #97, opened 2024-01-10
Development workflow in `./gen/nvml` 


msanft:
Hello there,

The README and the rules in the Makefile suggest the development of new bindings (when updating the header file) in the `gen` directory. On the other hand, the Go bindings are only created in the `pkg` directory, causing the `gen` directory, that hosts most of the wrapper code, to miss all of the definitions, mainly because `nvml.go` (i.e. the generation result) is not present there. Is there something I miss with the workflow, or would it just make a lot more sense to either have all of the definitions in `gen` and move everything excluding the gen-only files (i.e. `nvml.yml`) over on a release?

For now, I updated the bindings and the wrapper code in `pkg` only, which works. But if this should be upstreamed at some point, the whole repository and workflow should be kept in a good state probably.

Cheers, Moritz.

msanft:
Hey @elezar, thanks for the fast reply! 

Overall I understand why you want to have the in-development `gen` directory and the redistributable `pkg` folder. The issue I'm facing with this approach is that `nvml.go` is not present in the `gen` directory, what causes the Go language server (and builds in `gen`) to miss a lot of definitions. I think you should be able to reproduce this by performing a fresh clone and trying to build or use an editor with a Go language server in `gen`.

I think this could be fixed by also generating bindings in the `gen` directory, so that Go can find all definitions when writing code in `gen`. These can also be cleaned up after a build if you don't want these to clutter `gen`. I don't think this is an optimal approach, but I think it would ease development a little.

msanft:
Hey @elezar, this would be the correct solution imo, but also the one involving more changes. (as you said, makefile and readme need to be updated here)

Please let me know if I can help with #98, and thanks a ton for your efforts!

elezar:
> Please let me know if I can help with #98, and thanks a ton for your efforts!

If you want to take over #98 (it's probably simplest to create your own PR) and make the additional required, I'd be happy to review them and get them in.

PR #109, opened 2024-04-03
Move development directly into pkg/nvml instead of gen/nvml

klueska:
Fixes: https://github.com/NVIDIA/go-nvml/issues/97
Closes: https://github.com/NVIDIA/go-nvml/pull/98


Makefile
```diff
@@ -139,12 +139,11 @@ bindings: .create-bindings .strip-autogen-comment .strip-nvml-h-linenumber
 .create-bindings: $(PKG_BINDINGS_DIR)/nvml.h $(SOURCES) | $(PKG_BINDINGS_DIR)
 	cp $(GEN_BINDINGS_DIR)/nvml.yml $(PKG_BINDINGS_DIR)
 	c-for-go -out $(PKG_DIR) $(PKG_BINDINGS_DIR)/nvml.yml
```
elezar, line 141: Question: Instead of copying the `yml` file in the previous step, could we specify `$(GEN_BINDINGS_DIR)/nvml.yml` as input here? Then we also wouldn't need to remove it later?

elezar (approved):
One minor comment, but it's really not a blocker.

Makefile
klueska: When I do that I get an incorrect `pkg/types_gen.so` file that gets generated. It's unclear why this is the case, but the fact that it doesn't "just work" when simplifying this means that it likely warrants a new PR to investigate.

PR #110, opened 2024-04-03
Remove global newDynamicLibrary function and replace with embedded one

klueska:


pkg/nvml/lib.go
```diff
@@ -34,20 +34,27 @@ const (
...
+	dl                dynamicLibrary
+	newDynamicLibrary newDynamicLibraryFunc
 }
 
 // libnvml is a global instance of the nvml library.
 var libnvml = library{
 	path:  defaultNvmlLibraryName,
 	flags: defaultNvmlLibraryLoadFlags,
+	newDynamicLibrary: func(path string, flags int) dynamicLibrary {
+		return dl.New(path, flags)
```
elezar, line 56: the one thing that's strange here is that we're always calling this with `l.path` and `l.flags`. Are the arguments then needed at all?

elezar (approved):
Minor comment. Feel free to merge if it isn'tn relevant though.

pkg/nvml/lib.go
klueska: I actually had the same thought, but wasn't sure if maybe we would ever want a test that pulled in a library from a different path / with different flags. Independently, it seems nice to mimic what the actual `New()` call looks like and maybe we can actually push the ´dynamicLibrary` intertface down into the `dl` package some day so that this can turn into:
```
var libnvml = library{
	path:  defaultNvmlLibraryName,
	flags: defaultNvmlLibraryLoadFlags,
	newDynamicLibrary: dl.New
}
```

klueska: I played around with what you suggested and it felt more awkward than to leave it as is. I'll keep it in mind for the future though...

PR #111, opened 2024-04-04
Remove need for newDynamicLibraryFunc

klueska:
With the introduction of the refcount we no longer need to set library.dl to nil to track when it is loaded or unloaded. Instead we can use the refcount directly.


pkg/nvml/lib.go
```diff
@@ -282,20 +276,23 @@ func WithLibraryPath(path string) LibraryOption {
...
+	if o.path == "" {
+		o.path = defaultNvmlLibraryName
 	}
-	if libnvml.flags == 0 {
-		libnvml.flags = defaultNvmlLibraryLoadFlags
+	if o.flags == 0 {
+		o.flags = defaultNvmlLibraryLoadFlags
 	}
 
+	libnvml.dl = dl.New(o.path, o.flags)
```
elezar, line 295: Note that `libnvml` still has a `path` member. Should we set that to `o.path` here too?

klueska: Updated

elezar (approved):

PR #112, opened 2024-04-10
Wrap all opaque types as interfaces

klueska:


pkg/nvml/lib.go
```diff
@@ -149,9 +149,9 @@ var nvmlDeviceGetGridLicensableFeatures = nvmlDeviceGetGridLicensableFeatures_v1
 var nvmlEventSetWait = nvmlEventSetWait_v1
 var nvmlDeviceGetAttributes = nvmlDeviceGetAttributes_v1
 var nvmlComputeInstanceGetInfo = nvmlComputeInstanceGetInfo_v1
-var DeviceGetComputeRunningProcesses = deviceGetComputeRunningProcesses_v1
-var DeviceGetGraphicsRunningProcesses = deviceGetGraphicsRunningProcesses_v1
-var DeviceGetMPSComputeRunningProcesses = deviceGetMPSComputeRunningProcesses_v1
+var deviceGetComputeRunningProcesses = deviceGetComputeRunningProcesses_v1
```
elezar, line 152: Maybe a silly question: Can you remind me why we don't have an `nvml` prefix here?

pkg/nvml/api.go
```diff
@@ -349,6 +349,210 @@ type Interface interface {
 	GetVgpuDriverCapabilities(Capability VgpuDriverCapability) (bool, Return)
 }
 
+type Device interface {
+	GetName() (string, Return)
```
elezar, line 353: Can we sort these? (even if it is as a follow-up?)

pkg/nvml/gpm.go
```diff
@@ -44,23 +44,23 @@ func (l *library) GpmSampleAlloc(GpmSample *GpmSample) Return {
...
+	return nvmlGpmSampleGet(Device, GpmSample)
 }
 
 // nvml.GpmQueryDeviceSupport()
 type GpmSupportV struct {
-	device Device
+	device nvmlDevice
 }
 
 func (l *library) GpmQueryDeviceSupportV(Device Device) GpmSupportV {
```
elezar, line 69: This is probably out of scope, but does the `GpmQueryDeviceSupportV` top-level function (and others like it) add any value? Should we not *always* use the `GpmQueryDeviceSupportV` member function?

pkg/nvml/api.go
```diff
@@ -553,6 +553,11 @@ type Device interface {
 	VgpuTypeGetMaxInstances(VgpuTypeId VgpuTypeId) (int, Return)
 }
 
+type EventSet interface {
+	Wait(Timeoutms uint32) (EventData, Return)
```
elezar, line 575: `Timeoutms` -> `timeoutms`? (or just keep the types).

The case of the local variables is one of the biggest issues with enabling the linter in this package.

pkg/nvml/api.go
```diff
@@ -279,7 +279,7 @@ type Interface interface {
 	DeviceSetVgpuSchedulerState(Device Device, PSchedulerState *VgpuSchedulerSetState) Return
 	DeviceGetVgpuSchedulerCapabilities(Device Device) (VgpuSchedulerCapabilities, Return)
 	GpuInstanceGetComputeInstancePossiblePlacements(GpuInstance GpuInstance, Info *ComputeInstanceProfileInfo) ([]ComputeInstancePlacement, Return)
-	GpuInstanceCreateComputeInstanceWithPlacement(GpuInstance GpuInstance, Info *ComputeInstanceProfileInfo, Placement *ComputeInstancePlacement, ComputeInstance *ComputeInstance) Return
+	GpuInstanceCreateComputeInstanceWithPlacement(GpuInstance GpuInstance, Info *ComputeInstanceProfileInfo, Placement *ComputeInstancePlacement) (ComputeInstance, Return)
```
elezar, line 282: Is this a bug? Does it make sense to do this in a separate PR so as to not have it go "missing" in the changelog?

pkg/nvml/api.go
```diff
@@ -292,7 +292,7 @@ type Interface interface {
 	GpmMetricsGetV(MetricsGet *GpmMetricsGetType) GpmMetricsGetVType
 	GpmMetricsGet(MetricsGet *GpmMetricsGetType) Return
 	GpmSampleFree(GpmSample GpmSample) Return
-	GpmSampleAlloc(GpmSample *GpmSample) Return
+	GpmSampleAlloc() (GpmSample, Return)
```
elezar, line 295: Does it make sense to move this to a separate PR to call it out properly?

pkg/nvml/api.go
```diff
@@ -576,6 +576,12 @@ type EventSet interface {
 	Free() Return
 }
 
+type GpmSample interface {
+	Free() Return
+	Get(Device Device) Return
+	MigGet(Device Device, GpuInstanceId int) Return
```
elezar, line 582: Nit:
* `Device Device` -> `Device` or `device Device`
* `GpuInstanceId int` -> `int` or `gpuInstanceID int`.

elezar (commented):
Initial change look good. The main question I have is whether to top-level library functions that operate ON other types make sense, or whether we should always use the functions off the types?

pkg/nvml/lib.go
```diff
@@ -44,14 +44,37 @@ type library struct {
...
+	return newLibrary(opts...)
+}
+
+func newLibrary(opts ...LibraryOption) *library {
+	l := &library{}
+	l.init(opts...)
+	return l
+}
+
+func (l *library) init(opts ...LibraryOption) {
```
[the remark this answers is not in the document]
klueska: It's an internal type and only called from two places. One place holds the lock -- the other is when it is frist created so doesnt need to.

pkg/nvml/api.go
```diff
@@ -54,7 +49,597 @@ func SetLibraryOptions(opts ...LibraryOption) error {
...
+	GetVgpuCompatibility(nvmlVgpuMetadata *nvmlVgpuMetadata, PgpuMetadata *nvmlVgpuPgpuMetadata) (VgpuPgpuCompatibility, Return)
+	GetVgpuVersion() (VgpuVersion, VgpuVersion, Return)
+	SetVgpuVersion(VgpuVersion *VgpuVersion) Return
+	VgpuInstanceClearAccountingPids(VgpuInstance VgpuInstance) Return
+	VgpuInstanceGetMdevUUID(VgpuInstance VgpuInstance) (string, Return)
+	VgpuTypeGetCapabilities(VgpuTypeId VgpuTypeId, Capability VgpuCapability) (bool, Return)
+	GetVgpuDriverCapabilities(Capability VgpuDriverCapability) (bool, Return)
 }
+
+// Define package level methods as aliases to Interface methods of libnvml
```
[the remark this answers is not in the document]
klueska: Not at the moment. It was fairly easy to do though. I just grepped for `func (l* library)` copied that in and did some search and replace on it.

pkg/nvml/lib.go
klueska: The nvml "functions" in this block come from #defines in nvml.h.  The underlying versioned functions that get assigned to these variables are generated as part of the bindings with the `v1`, `v2`, `v3`, etc. prefixes and then assigned in `updateVersionedSymbols` below.

The difference with these functions is that we need to wrap the underlying `nvml*_v1`, `nvml*_v2`, etc. functions with some extra, version specific logic. Because of this, we create a level of indirection before assinging them back out to the "unversioned" variable.

pkg/nvml/gpm.go
klueska: No, you don't need to. If you just want the latest, you call e.g. `GpmQueryDeviceSupport()` directly. Only if you care about a specific version do you call `GpmQueryDeviceSupportV().V1()`, `GpmQueryDeviceSupportV().V2()`, etc.

pkg/nvml/api.go
klueska: I am doing a pass now to lowercase all of the variable names. It was causing problems with `moq` as well. It's the reason this is still a draft PR until I get these updated.

pkg/nvml/api.go
klueska: Yes, it is a bug. However, if I do it as a separate PR, then I can't make the updates in the commit that follows to turn ComputeInstance into an Interface. I'd rather have all the interface conversion in a single PR. I could look at maybe updating it in a "pre" PR, but I don't know how hard it will be to rebase if I do that.

pkg/nvml/api.go
klueska: Ditto here, I will look into it as a "pre" PR.

pkg/nvml/api.go
klueska: Yes, I am doing a pass now to lowercase all variable names.

pkg/nvml/api.go
elezar: It's not a blocker -- as long as we actually call it out in the release notes.

pkg/nvml/api.go
elezar: We can do it as a follow-up if that gets too much.

klueska: I have made this change, updated, and moved the PR out of draft.

klueska:
> Initial change look good. The main question I have is whether to top-level library functions that operate ON other types make sense, or whether we should always use the functions off the types?

I'd be fine returning concrete types for everything (including the top level library) and moving the API definitions to the `testing` package. I went back and forth on this and don't really have a preference, so long as we have a way to pull in a mock version of the library (and its other types) from somewhere.

elezar:
> > Initial change look good. The main question I have is whether to top-level library functions that operate ON other types make sense, or whether we should always use the functions off the types?

No, that's not what I meant -- although it is a valid question. I meant that we have, as an example, three functions that do the same thing:
```
DeviceGetComputeMode(device Device)  (ComputeMode, Return)
library.DeviceGetComputeMode(device Device) (ComputeMode, Return)
Device.GetComputeMode()  (ComputeMode, Return)
```
My question was whether we would consider moving towards requiring the use of `Device.GetComputeMode()` and at least deprecate the other usages? This would drastically reduce the size of the top-level `Interface`.

klueska:
> > > Initial change look good. The main question I have is whether to top-level library functions that operate ON other types make sense, or whether we should always use the functions off the types?
> 
> No, that's not what I meant -- although it is a valid question. I meant that we have, as an example, three functions that do the same thing:
> 
> ```
> DeviceGetComputeMode(device Device)  (ComputeMode, Return)
> library.DeviceGetComputeMode(device Device) (ComputeMode, Return)
> Device.GetComputeMode()  (ComputeMode, Return)
> ```
> 
> My question was whether we would consider moving towards requiring the use of `Device.GetComputeMode()` and at least deprecate the other usages? This would drastically reduce the size of the top-level `Interface`.

I would be inclined to keep them since that is what matches the actual NVML API. That said, we should add tooling to autogenerate the top level Interface as well as the package level aliases from the functions hanging off the `library` type.

pkg/nvml/api.go
elezar: OK. Ordering the same as `nvml.h` is good -- this should then aid automation and also make diffs easy enough to read. No need for this follow up then.

pkg/nvml/api.go
elezar: I'm happy to leave this in this PR as long as we remember to call it out in the release notes once we tag a new version.

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,388 @@
+package main
```
elezar, line 1: License?

pkg/nvml/zz_generated.api.go
```diff
@@ -0,0 +1,885 @@
+// Generated Code; DO NOT EDIT.
```
elezar, line 1: Do we still want to add the license boilerplate?

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,388 @@
...
+		fmt.Fprintf(writer, output)
+
+		if i < (len(GeneratableInterfaces) - 1) {
+			fmt.Fprintf(writer, "\n")
+		}
+	}
+}
+
+func getWriter(outputFile string) (io.Writer, func() error, error) {
+	if outputFile == "-" {
```
elezar, line 132: Is `-` not used for `STDIN`? and `""` for `STDOUT`? (this is a nit).

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,388 @@
...
+		return os.Stdout, nil, nil
```
elezar, line 133: Instead of `nil` as the "Closer" can we return:
```
func() error { return nil }
```
then we don't need to check for nil at the close site?

Alternatively, we can return a `io.WriteCloser` and implement a `noopCloser` that we combine with os.Stdout:
```
type noopCloser struct {
   io.Writer
}

func (n *noopCloser) Close() error {
   return nil
}
```
And then our return statement becomes:
```
return &noopCloser{Writer: os.Stdout}, nil
```
(and we can just `return os.Create(outputFile)` directly in the file case).

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,388 @@
...
+		if err != nil {
+			return err
+		}
+
+		gofiles[path] = content
+
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("walking the embed.FS path: %w\n", err)
```
elezar, line 243: embed.FS?

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+
+		comment, err := generateInterfaceComment(p)
+		if err != nil {
+			fmt.Printf("Error: %v", err)
+			return
+		}
+		fmt.Fprintf(writer, comment)
+
+		output, err := generateInterface(*sourceDir, p)
```
elezar, line 118: Do we fmt the output automaticallty?

elezar (commented):
Some minor comments on the generation toolking to start with.

gen/nvml/generateapi.go
klueska: Yes, you are right -- will fix

gen/nvml/generateapi.go
klueska: I went with the first suggestion

klueska: As you know, I am not a fan of "noop code" and prefer to be explicit (which is why I hesitated even changing this at all, but this is minor, so its fine.

gen/nvml/generateapi.go
klueska: Leftover from an old iteration, removing

gen/nvml/generateapi.go
klueska: I now call `make fmt` as the last step in `make bindings` where the call to run this script is invoked.

pkg/nvml/api.go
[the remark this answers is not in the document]
klueska: The tool has been added in the latest revision.

gen/nvml/generateapi.go
klueska: Done

gen/nvml/generateapi.go
klueska: Done

pkg/nvml/zz_generated.api.go
klueska: Added in latest

gen/nvml/generateapi.go
klueska: Added in latest

go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/NVIDIA/go-nvml
 
 go 1.20
 
-require github.com/stretchr/testify v1.9.0
+require (
+	github.com/google/uuid v1.6.0
```
elezar, line 6: As a matter of interest, where is this dependency from?

gen/nvml/generateapi.go
```diff
@@ -0,0 +1,420 @@
...
+		return nil, nil, err
+	}
+
+	return file, file.Close, nil
+}
+
+func generateHeader() (string, error) {
+	lines := []string{
+		"/**",
+		"# Copyright 2024 NVIDIA CORPORATION",
```
elezar, line 161: As a follow up we may want to read a `boilerplate.txt` from disk instead so as to not have to update the date.

pkg/nvml/mock/dgxa100.go
elezar, line 1: As a thought. Does it make sense to have the "implementations" under something like `mock/system` or `mock/platform`?

elezar (dismissed):
Minor comment about where we put our Mock sytems implementations.

go.mod
klueska: It's to generate a UUID for the GPUs in the mock dgxa100 server

pkg/nvml/mock/dgxa100.go
klueska: I'd be happy with a `dgxa100` folder though. Then we could change the names to `dgxa100.Server` and `dgxa100.Device`, `dgxa100.GpuInstance`, etc.

klueska: I made this change -- let me know what you think

go.mod
elezar: Ah. OK. Thanks.

pkg/nvml/mock/dgxa100.go
elezar: > What value does the extra level of nesting bring? 
It's more about organisation than anything else. The mocks themselves are all generated wherease the systems are implemented.

pkg/nvml/mock/dgxa100/dgxa100.go
```diff
@@ -0,0 +1,565 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
elezar, line 2: Nit: The date.

elezar (dismissed):
Thanks for this @klueska.

I have some local changes in the device plugin that could benefit from the "Mock" implementations already!

elezar (dismissed):

pkg/nvml/lib.go
```diff
@@ -85,6 +113,10 @@ func (l *library) load() (rerr error) {
 		return fmt.Errorf("error opening %s: %w", l.path, err)
 	}
 
+	// Update the errorStringFunc to point to nvml.ErrorString
+	errorStringFunc = nvmlErrorString
```
elezar, line 117: Do we need to set this to the default once we close the library?

pkg/nvml/return.go
```diff
@@ -19,17 +19,27 @@ import (
 )
 
 // nvml.ErrorString()
-func ErrorString(r Return) string {
-	if err := GetLibrary().Lookup("nvmlErrorString"); err != nil {
-		return fallbackErrorStringFunc(r)
-	}
-	return nvmlErrorString(r)
+func (l *library) ErrorString(r Return) string {
+	return r.Error()
```
elezar, line 23: We either need to keep the lookup logic as before or reset the default if the library is unloaded.

elezar (commented):
Thanks. Looks good.

Minor comment regarding the error string.

pkg/nvml/lib.go
```diff
@@ -43,27 +53,45 @@ type library struct {
...
+	}
+	if o.flags == 0 {
+		o.flags = defaultNvmlLibraryLoadFlags
+	}
+
+	l.path = o.path
+	l.dl = dl.New(o.path, o.flags)
+}
+
+func (l *library) GetExtendedInterface() ExtendedInterface {
```
elezar, line 88: I'm not sold on `GetExtendedInterface` and would prefer `GetExtensions()`, but this is not a blocker. I'm happy to keep it as is and revisit this later once we have some usage examples.

pkg/nvml/return.go
klueska: I can't leave it as before because in the r.Error() and r.String() methods I don't have access to `l`. I will update it reset the function on unload.

pkg/nvml/lib.go
klueska: Updated as suggested.

klueska: Actually, I went for a hybrid -- the interface is called `ExtendedInterface` to be symmetrical with `Interface`, but the method call is `Extensions()` (which is the only actual symbol anyone will see in practice), e.g.:
```
nvml.Extensions().LookupSymbol("symbol")
```
or
```
envml := nvml.Extensions()
envml.LookupSymbol("symbol")
```

Its possible that the type itself might be embedded in a struct, but I only see that happening in testing (if at all) and I like the symmetry of that with `Interface` for the core API.

pkg/nvml/return.go
klueska: Updated.

pkg/nvml/lib.go
klueska: Yes. Fixed.

pkg/nvml/lib.go
elezar: That sounds fine. I'm not too concerned about the type name, since as you mention that isn't visible at the call site.

elezar (approved):
Looks good now!

PR #113, opened 2024-04-22
Fix bug with wrong instance of lib being called for load/close

klueska:
Without this change, go-nvlib is broken, when looking up symbol names, for example. Also, no v2 or v3 symbols are loaded as expected which breaks, for example, `GetGpuInstancePossiblePlacements()` for the extended media types.


elezar (approved):
Does this mean we need to update `mig-parted` to include this fix?

Are there other functions that are critical for our own components?

PR #114, opened 2024-04-23
Add additional functionality to the mock dgxa100 server

klueska:


pkg/nvml/mock/dgxa100/dgxa100.go
```diff
@@ -26,325 +26,49 @@ import (
...
-				SharedDecoderCount:    5,
-				SharedEncoderCount:    5,
-				SharedJpegCount:       1,
-				SharedOfaCount:        1,
-			},
-		},
-	},
-}
-
 func New() nvml.Interface {
```
elezar, line 72: Question: Does using the generated mocks add value here? 

It means that we don't have to implement an exhaustive interface and can rely on the stubbing by `moq` (or allow it to panic if an unimplemneted function is called).

pkg/nvml/mock/dgxa100/dgxa100.go
```diff
@@ -357,14 +81,25 @@ func New() nvml.Interface {
...
 	}
 }
 
 func NewDevice(index int) nvml.Device {
 	return &Device{
-		UUID:               "GPU-" + uuid.New().String(),
-		PciBusID:           fmt.Sprintf("0000:%02x:00.0", index),
-		Index:              index,
+		UUID:         "GPU-" + uuid.New().String(),
+		Name:         "NVIDIA A100-SXM4-40GB",
```
elezar, line 93: Do we want to add `Mock` to the name?

pkg/nvml/mock/dgxa100/dgxa100.go
```diff
@@ -26,325 +26,49 @@ import (
 
 type Server struct {
 	mock.Interface
-	Devices [8]nvml.Device
+	Devices           [8]nvml.Device
+	DriverVersion     string
+	NvmlVersion       string
```
elezar, line 31: Question: When is the `NvmlVersion` relevant?

elezar (commented):
Just a question around implementing an explicit mock vs using the generated mocks.

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: I'm not sure I follow. The types in this packag extend the generated mock type for exactly this purpose. I only overwrite the methods I care about, and the rest are left as stubs that panic.

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: Seems reasonable.

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: Not sure exactly, but the code already had an API call for it, so I updated it to pull from the Server field now instead of hard-coding it in the method call. I assume mig-parted must rely on it since that is where this mock server was originally housed.

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: It actually how figured out which ones I needed to implement -- those that were missing were panicking via the moq stubs.

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: Updated.

pkg/nvml/mock/dgxa100/dgxa100.go
elezar: One thing to note is that this breaks information that we get from the mock such as the call history -- number of calls and the parameters.

The "Correct" way to handle this is to implement the funcs as required. See for example: https://github.com/NVIDIA/go-gpuallocator/blob/82d8924b3d40c7d31cc18d56e65b3bffe59d91fd/gpuallocator/device_test.go#L28-L52

pkg/nvml/mock/dgxa100/dgxa100.go
```diff
@@ -357,14 +81,25 @@ func New() nvml.Interface {
 			NewDevice(6),
 			NewDevice(7),
 		},
+		DriverVersion:     "550.54.15",
+		NvmlVersion:       "12.550.54.15",
+		CudaDriverVersion: 12040,
```
elezar, line 86:
```suggestion
		SystemGetDriverVersionFunc: func() (string, nvml.Return) {
		    return "550.54.15", nvml.SUCCESS
		},
		SystemGetNvmlVersionFunc: func() (string, nvml.Return) {
		      return "12.550.54.15", nvml.SUCCESS
		 },
		SystemGetCudaDriverVersionFunc: func() (int, nvml.Return) {
		     return 12040, nvml.SUCCESS
		},
```

pkg/nvml/mock/dgxa100/dgxa100.go
klueska: Good point. Updated throughout.

elezar (dismissed):
Looks good. Let's get these in and iterate if we need anything else.

One thought ... if we return mock instances from the constructors instead of `nvml.Interface` and `nvml.Device` a caller can add / override mocks as required without having to modify the "upstream" code.

Probably out of scope for this PR though.

klueska:
No, I think that makes sense. It will still implement the interface (so we can assign it to the interface in the calling code), but it can also be extended if desired from the calling code (which the current implementation doesn't allow). Let me add that in this PR.

klueska:
OK, updated with that change. I also vendored it into mig-parted and tested it to make sure it didn't break anything by doing this.

elezar (dismissed):
LGTM.

klueska:
Added one more commit to introduce locks to protect from concurrent reads/writes of maps

elezar (approved):
lgtm

PR #119, opened 2024-05-08
Add internal method to get device handle

elezar:
This was done in go-nvlib in https://github.com/NVIDIA/go-nvlib/pull/17 but not included when we implemented the interfaces here.

Without the explict `nvmlDevice` conversion we see:
```
=== RUN   TestGetTopologyCommonAncestor
--- FAIL: TestGetTopologyCommonAncestor (0.00s)
panic: interface conversion: nvml.Device is nvml.wrappedDevice, not nvml.nvmlDevice [recovered]
        panic: interface conversion: nvml.Device is nvml.wrappedDevice, not nvml.nvmlDevice

goroutine 19 [running]:
testing.tRunner.func1.2({0x1030fc500, 0xc0000b10b0})
        /usr/local/Cellar/go/1.22.2/libexec/src/testing/testing.go:1631 +0x24a
testing.tRunner.func1()
        /usr/local/Cellar/go/1.22.2/libexec/src/testing/testing.go:1634 +0x377
panic({0x1030fc500?, 0xc0000b10b0?})
        /usr/local/Cellar/go/1.22.2/libexec/src/runtime/panic.go:770 +0x132
github.com/NVIDIA/go-nvml/pkg/nvml.nvmlDevice.GetTopologyCommonAncestor(...)
        /Users/elezar/src/go-nvml/pkg/nvml/device.go:223
github.com/NVIDIA/go-nvml/pkg/nvml.TestGetTopologyCommonAncestor(0xc0000c29c0)
        /Users/elezar/src/go-nvml/pkg/nvml/device_test.go:38 +0x85
testing.tRunner(0xc0000c29c0, 0x10313b4c8)
        /usr/local/Cellar/go/1.22.2/libexec/src/testing/testing.go:1689 +0xfb
created by testing.(*T).Run in goroutine 1
        /usr/local/Cellar/go/1.22.2/libexec/src/testing/testing.go:1742 +0x390
FAIL    github.com/NVIDIA/go-nvml/pkg/nvml      0.612s
```

### DIFF
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -220,9 +220,12 @@ func (l *library) DeviceGetTopologyCommonAncestor(device1 Device, device2 Device
 
 func (device1 nvmlDevice) GetTopologyCommonAncestor(device2 Device) (GpuTopologyLevel, Return) {
 	var pathInfo GpuTopologyLevel
-	ret := nvmlDeviceGetTopologyCommonAncestor(device1, device2.(nvmlDevice), &pathInfo)
+	ret := nvmlDeviceGetTopologyCommonAncestorStub(device1, device2.(nvmlDevice), &pathInfo)
 	return pathInfo, ret
 }
+
+// nvmlDeviceGetTopologyCommonAncestorStub allows us to override this for testing.
+var nvmlDeviceGetTopologyCommonAncestorStub = nvmlDeviceGetTopologyCommonAncestor
 
 // nvml.DeviceGetTopologyNearestGpus()
 func (l *library) DeviceGetTopologyNearestGpus(device Device, level GpuTopologyLevel) ([]Device, Return) {
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -15,8 +15,47 @@
 package nvml
 
 import (
+	"fmt"
+	"reflect"
 	"unsafe"
 )
+
+// nvmlDeviceHandle attempts to convert a device d to an nvmlDevice.
+// This is required for functions such as GetTopologyCommonAncestor which
+// accept Device arguments that need to be passed to internal nvml* functions
+// as nvmlDevice parameters.
+func nvmlDeviceHandle(d Device) nvmlDevice {
+	var helper func(val reflect.Value) nvmlDevice
+	helper = func(val reflect.Value) nvmlDevice {
+		if val.Kind() == reflect.Interface {
+			val = val.Elem()
+		}
+
+		if val.Kind() == reflect.Ptr {
+			val = val.Elem()
+		}
+
+		if val.Type() == reflect.TypeOf(nvmlDevice{}) {
+			return val.Interface().(nvmlDevice)
+		}
+
+		if val.Kind() != reflect.Struct {
+			panic(fmt.Errorf("unable to convert non-struct type %v to nvmlDevice", val.Kind()))
+		}
+
+		for i := 0; i < val.Type().NumField(); i++ {
+			if !val.Type().Field(i).Anonymous {
+				continue
+			}
+			if !val.Field(i).Type().Implements(reflect.TypeOf((*Device)(nil)).Elem()) {
+				continue
+			}
+			return helper(val.Field(i))
+		}
+		panic(fmt.Errorf("unable to convert %T to nvmlDevice", d))
+	}
+	return helper(reflect.ValueOf(d))
+}
 
 // EccBitType
 type EccBitType = MemoryErrorType
@@ -220,7 +259,7 @@ func (l *library) DeviceGetTopologyCommonAncestor(device1 Device, device2 Device
 
 func (device1 nvmlDevice) GetTopologyCommonAncestor(device2 Device) (GpuTopologyLevel, Return) {
 	var pathInfo GpuTopologyLevel
-	ret := nvmlDeviceGetTopologyCommonAncestorStub(device1, device2.(nvmlDevice), &pathInfo)
+	ret := nvmlDeviceGetTopologyCommonAncestorStub(device1, nvmlDeviceHandle(device2), &pathInfo)
 	return pathInfo, ret
 }
 
@@ -253,7 +292,7 @@ func (l *library) DeviceGetP2PStatus(device1 Device, device2 Device, p2pIndex Gp
 
 func (device1 nvmlDevice) GetP2PStatus(device2 Device, p2pIndex GpuP2PCapsIndex) (GpuP2PStatus, Return) {
 	var p2pStatus GpuP2PStatus
-	ret := nvmlDeviceGetP2PStatus(device1, device2.(nvmlDevice), p2pIndex, &p2pStatus)
+	ret := nvmlDeviceGetP2PStatus(device1, nvmlDeviceHandle(device2), p2pIndex, &p2pStatus)
 	return p2pStatus, ret
 }
 
@@ -1185,7 +1224,7 @@ func (l *library) DeviceOnSameBoard(device1 Device, device2 Device) (int, Return
 
 func (device1 nvmlDevice) OnSameBoard(device2 Device) (int, Return) {
 	var onSameBoard int32
-	ret := nvmlDeviceOnSameBoard(device1, device2.(nvmlDevice), &onSameBoard)
+	ret := nvmlDeviceOnSameBoard(device1, nvmlDeviceHandle(device2), &onSameBoard)
 	return int(onSameBoard), ret
 }
 
--- a/pkg/nvml/device_test.go
+++ b/pkg/nvml/device_test.go
@@ -27,18 +27,64 @@ func TestGetTopologyCommonAncestor(t *testing.T) {
 		Device
 	}
 
-	d1 := wrappedDevice{
-		Device: nvmlDevice{},
+	type wrappedWrappedDevice struct {
+		wrappedDevice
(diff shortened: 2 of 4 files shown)


klueska:
I don't think this approach will work in general. It works for new types that embed the `Device` type, but not for those that want to implement the `Device` type from scratch (e.g. our mock implementation) because they have no way of providing an implementation for the unexported method.

Would an alternative be to implement the following internal function directly? I realize it relies on reflection, but maybe this is OK since it won't actually be called very often...

```
func nvmlDeviceHandle(d Device) nvmlDevice {
	if nvmlDevice, ok := d.(nvmlDevice); ok {
		return nvmlDevice
	}
	
	val := reflect.ValueOf(d)
	typ := val.Type()
	for i := 0; i < typ.NumField(); i++ {
		if !typ.Field(i).Anonymous {
			continue
		}
		if _, ok := val.Field(i).Interface().(Device); !ok {
			continue
		}
		return nvmlDeviceHandle(val.Field(i).Interface().(Device))
	}
	panic("Unable to convert Device to underlying nvmlDevice")
}
```

Here it is in the go playground:
https://go.dev/play/p/AnJXQpCrzGZ?v=gotip.go?download=true.go?download=true

elezar:
There is definitely a concern that this would not work in general. I also considered reflect, but didn't have time to properly flesh this out.

Note that the issue here is not the concrete type to which the function is attached, but the fact that we're accepting an interface type for one of the functions and expect this argument to be convertable to the base type because we need to call an internal function with this type. It is this call to the internal function that causes a panic when the input cannot be converted to the concrete type.

I have incorportated your `nvmlDeviceHandle` function in the latest revision and we will still panic in cases where we cannot perform this conversion, but at least cover the wrapped cases. For users implementing their own `Device` types from scratch, they will have to update the relevant functions accordingly.

Another option would be to add a function that provides the underlying reference as a public function. This may still not make sense in mocks, but would allow consumers to implement this as they require.

pkg/nvml/device.go
```diff
@@ -15,9 +15,38 @@
...
+// as nvmlDevice parameters.
+func nvmlDeviceHandle(d Device) nvmlDevice {
+	if nvmlDevice, ok := d.(nvmlDevice); ok {
+		return nvmlDevice
+	}
+
+	val := reflect.ValueOf(d)
+	typ := val.Type()
+
+	if typ.Kind() == reflect.Pointer {
```
elezar, line 35: Is required for:
```
panic: reflect: NumField of non-struct type *device.device
```

klueska: Instead of this, can we replace `val := reflect.ValueOf(d)` above with:
```
	val := reflect.ValueOf(d)
	for val.Kind() == reflect.Ptr {
		val = val.Elem()
	}
```

klueska: Do we also need to check to see if it is a struct type and panic with a more convenient error if it is not?

elezar: I added an explicit check. Not sure if we add more value, but at least we have a point where we can extend the type checks.

As a matter of interest, why did you use `for` above instead of `if`?

pkg/nvml/device.go
```diff
@@ -15,9 +15,42 @@
...
+// This is required for functions such as GetTopologyCommonAncestor which
+// accept Device arguments that need to be passed to internal nvml* functions
+// as nvmlDevice parameters.
+func nvmlDeviceHandle(d Device) nvmlDevice {
+	if nvmlDevice, ok := d.(nvmlDevice); ok {
+		return nvmlDevice
+	}
+
+	val := reflect.ValueOf(d)
+	if val.Kind() == reflect.Ptr {
```
klueska, line 33: I had used a for loop here to cover cases where we have a pointer to a pointer?
```suggestion
	for val.Kind() == reflect.Ptr {
```

elezar: OK. That's fair. Could you also provide an example for the unit test?

klueska: I can't seem to come up with one very easily. It may even be that since the interface is defined on `nvmlDevice` and not `*nvmlDevice` that it's actually impossible to construct a use-case where this would be needed. Feel free to ignore it if you don' think it's necessary. It likely won't come up in practice.

klueska: However, this legitimate/real use-case breaks the current implementation:
```
diff --git a/pkg/nvml/device_test.go b/pkg/nvml/device_test.go
index 9dd4c07..a4b9ea3 100644
--- a/pkg/nvml/device_test.go
+++ b/pkg/nvml/device_test.go
@@ -27,6 +27,10 @@ func TestGetTopologyCommonAncestor(t *testing.T) {
        Device
    }

+   type wrappedDevice2 struct {
+       wrappedDevice
+   }
+
    testCases := []struct {
        description string
        device      Device
@@ -43,6 +47,17 @@ func TestGetTopologyCommonAncestor(t *testing.T) {
                Device: nvmlDevice{},
            },
        },
+       {
+           description: "pointer to a pointer of a wrapped device",
+           device: func() Device {
+               d := &wrappedDevice2{
+                   wrappedDevice: wrappedDevice{
+                       Device: nvmlDevice{},
+                   },
+               }
+               return d
+           }(),
+       },
    }

    for _, tc := range testCases {
```

klueska:
```
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method [recovered]
	panic: reflect.Value.Interface: cannot return value obtained from unexported field or method
```

klueska: This seems to fix that:
```
diff --git a/pkg/nvml/device.go b/pkg/nvml/device.go
index 8daeedc..8a871d5 100644
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -43,10 +43,21 @@ func nvmlDeviceHandle(d Device) nvmlDevice {
 		if !typ.Field(i).Anonymous {
 			continue
 		}
-		if _, ok := val.Field(i).Interface().(Device); !ok {
+
+		// Make a copy of the original value to ensure that it (as well as all
+		// of its embedded fields) are addressable.
+		addressableVal := reflect.New(val.Type()).Elem()
+		addressableVal.Set(val)
+
+		// Create a copy of the variable at val.Field(i) that we know we can access
+		// (even if its an unexported type).
+		field := addressableVal.Field(i)
+		field = reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem()
+
+		if _, ok := field.Interface().(Device); !ok {
 			continue
 		}
-		return nvmlDeviceHandle(val.Field(i).Interface().(Device))
+		return nvmlDeviceHandle(field.Interface().(Device))
 	}
 	panic(fmt.Errorf("unable to convert %T to nvmlDevice", d))
 }
diff --git a/pkg/nvml/device_test.go b/pkg/nvml/device_test.go
index 9dd4c07..bf7f09a 100644
--- a/pkg/nvml/device_test.go
+++ b/pkg/nvml/device_test.go
@@ -27,6 +27,10 @@ func TestGetTopologyCommonAncestor(t *testing.T) {
 		Device
 	}

+	type wrappedDevice2 struct {
+		wrappedDevice
+	}
+
 	testCases := []struct {
 		description string
 		device      Device
@@ -43,6 +47,14 @@ func TestGetTopologyCommonAncestor(t *testing.T) {
 				Device: nvmlDevice{},
 			},
 		},
+		{
+			description: "nested embedded device with concrete type",
+			device: wrappedDevice2{
+				wrappedDevice: wrappedDevice{
+					Device: nvmlDevice{},
+				},
+			},
+		},
 	}

 	for _, tc := range testCases {
```

klueska: Or maybe this alternative implementation is clearer:
```
func nvmlDeviceHandle(d Device) nvmlDevice {
    var helper func(val reflect.Value) nvmlDevice
    helper = func(val reflect.Value) nvmlDevice {
        if val.Kind() == reflect.Interface {
            val = val.Elem()
        }

        if val.Kind() == reflect.Ptr {
            val = val.Elem()
        }

        if val.Type() == reflect.TypeOf(nvmlDevice{}) {
            return val.Interface().(nvmlDevice)
        }

        if val.Kind() != reflect.Struct {
            panic(fmt.Errorf("unable to convert non-struct type %v to nvmlDevice", val.Kind()))
        }

        for i := 0; i < val.Type().NumField(); i++ {
            if !val.Type().Field(i).Anonymous {
                continue
            }
            if !val.Field(i).Type().Implements(reflect.TypeOf((*Device)(nil)).Elem()) {
                continue
            }
            return helper(val.Field(i))
        }
        panic(fmt.Errorf("unable to convert %T to nvmlDevice", d))
    }
    return helper(reflect.ValueOf(d))
}
```

elezar: I like the helper-based implementation more -- although it's not clear what the benefits are over just calling the top-level function recursively instead of calling `helper` recursively.

Have updated and added some more tests.

klueska: The problem is that we can't call `Interface()` on a non-exported field to get its concrete type (which is the case covered by my new test). And without the ability to do that, we can't call the top-level function since it takes a concrete type (as opposed to a `reflect.Value`).

klueska:
I'm still sad there isn't a good way to do this without reflection, but I think the current implementation is pretty easy to understand and covers all our bases.

klueska (approved):

PR #122, opened 2024-05-16
Fix missing GPM metrics

### LINKED ISSUE #121: Collected GPM metrics are not returned to caller
When calling `GpmMetricsGet` we convert the `GpmMetricsGetType` to an `nvmlGpmMetricsGetType` for the call to `nvmlGpmMetricsGet`.

On returning from `nvmlGpmMetricsGet` the contents of the `nvmlGpmMetricsGetType.Metrics` field corresponds to the collected metrics. These are never transferred into the `GpmMetricsGetType` passed as input.

This should be addressed by changing:
```
func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
	metricsGet.Version = GPM_METRICS_GET_VERSION
	return nvmlGpmMetricsGet(metricsGet.convert())
}
```

to something equivalent to:
```
func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
	metricsGet.Version = GPM_METRICS_GET_VERSION
	nvmlMetricsGet := metricsGet.convert()
	defer func() {
		for i := range nvmlMetricsGet.Metrics {
			metricsGet.Metrics[i] = nvmlMetricsGet.Metrics[i]
		}
	}()
	return nvmlGpmMetricsGet(nvmlMetricsGet)
}
```

elezar:
These changes ensure that the collected GPM metrics are transferred to the original data structure passed to `GpmMetricsGet`.

Closes #121

### DIFF
--- a/pkg/nvml/gpm.go
+++ b/pkg/nvml/gpm.go
@@ -51,20 +51,31 @@ func (g *nvmlGpmMetricsGetType) convert() *GpmMetricsGetType {
 
 // nvml.GpmMetricsGet()
 type GpmMetricsGetVType struct {
-	metricsGet *nvmlGpmMetricsGetType
+	metricsGet *GpmMetricsGetType
 }
 
 func (l *library) GpmMetricsGetV(metricsGet *GpmMetricsGetType) GpmMetricsGetVType {
-	return GpmMetricsGetVType{metricsGet.convert()}
+	return GpmMetricsGetVType{metricsGet}
 }
+
+// nvmlGpmMetricsGetStub is a stub function that can be overridden for testing.
+var nvmlGpmMetricsGetStub = nvmlGpmMetricsGet
+
 func (metricsGetV GpmMetricsGetVType) V1() Return {
 	metricsGetV.metricsGet.Version = 1
-	return nvmlGpmMetricsGet(metricsGetV.metricsGet)
+	return gpmMetricsGet(metricsGetV.metricsGet)
 }
 
 func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
 	metricsGet.Version = GPM_METRICS_GET_VERSION
-	return nvmlGpmMetricsGet(metricsGet.convert())
+	return gpmMetricsGet(metricsGet)
+}
+
+func gpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
+	nvmlMetricsGet := metricsGet.convert()
+	ret := nvmlGpmMetricsGetStub(nvmlMetricsGet)
+	*metricsGet = *nvmlMetricsGet.convert()
+	return ret
 }
 
 // nvml.GpmSampleFree()
--- a/examples/gpm-metrics/main.go
+++ b/examples/gpm-metrics/main.go
@@ -1,0 +1,111 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+func main() {
+	ret := nvml.Init()
+	if ret != nvml.SUCCESS {
+		log.Fatalf("failed to init NVML: %v", ret)
+	}
+	defer func() {
+		_ = nvml.Shutdown()
+	}()
+
+	count, ret := nvml.DeviceGetCount()
+	if ret != nvml.SUCCESS {
+		log.Fatalf("failed to get device count: %v", ret)
+	}
+
+	for i := 0; i < count; i++ {
+		if err := collectGPMMetrics(i); err != nil {
+			log.Printf("failed to get metrics for device %d: %v\n", i, err)
+		}
+	}
+}
+
+// collectGPMMetrics gets GPM metrics for a specified device.
+func collectGPMMetrics(i int) error {
+	device, ret := nvml.DeviceGetHandleByIndex(i)
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("could not get devices handle: %w", ret)
+	}
+
+	gpuQuerySupport, ret := device.GpmQueryDeviceSupport()
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("could not query GPM support: %w", ret)
+	}
+
+	if gpuQuerySupport.IsSupportedDevice == 0 {
+		return fmt.Errorf("GPM queries are not supported")
+	}
+
+	sample1, ret := nvml.GpmSampleAlloc()
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("could not allocate sample: %w", ret)
+	}
+	defer func() {
+		_ = sample1.Free()
+	}()
+	sample2, ret := nvml.GpmSampleAlloc()
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("could not allocate sample: %w", ret)
+	}
+	defer func() {
+		_ = sample2.Free()
+	}()
+
+	if ret := device.GpmSampleGet(sample1); ret != nvml.SUCCESS {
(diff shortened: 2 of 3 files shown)


pkg/nvml/gpm.go
```diff
@@ -57,14 +57,23 @@ type GpmMetricsGetVType struct {
...
+	return nvmlGpmMetricsGetStub(metricsGetV.metricsGet)
 }
 
 func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
 	metricsGet.Version = GPM_METRICS_GET_VERSION
-	return nvmlGpmMetricsGet(metricsGet.convert())
+	nvmlMetricsGet := metricsGet.convert()
+	defer func() {
+		metricsGet.Metrics = nvmlMetricsGet.Metrics
+	}()
```
klueska, line 74: Can we avoid doing a defer like this and do something more explicit?

klueska: I remember now -- because the API requires some fields to be set as INPUTs and other fields to be set as OUTPUTs in the struct: https://github.com/NVIDIA/go-nvml/blob/main/gen/nvml/nvml.h#L9401-L9409

pkg/nvml/gpm.go
```diff
@@ -57,14 +57,23 @@ type GpmMetricsGetVType struct {
...
+
+	return nvmlGpmMetricsGetStub(nvmlMetricsGet)
```
klueska, line 76: I don't like the idea of adding a one-off stub like this. If we are going to do this, we should do it for everything.

elezar: To be clear, I added one like this for the topology functions that I also added tests for. I didn't see the value in adding all of them if we weren't going to use them immediately.

Would you rather I remove the tests and the stubs?

klueska: I guess it's fine then. I think we can add a level of indirection pretty easily though. We just instruct `nvidia.yml` to rewrite all generated functions as `nvml*Internal` and then do something similar to how I generate the various `interfaces` to generate variables for all `nvml* = nvml*Internal` by default. Even better would be to make them all function pointers hanging off of the `library` type (so we don't have to do the weird update/reset in the tests), but that might be more trouble than it's worth.

pkg/nvml/gpm.go
```diff
@@ -57,14 +57,23 @@ type GpmMetricsGetVType struct {
 func (l *library) GpmMetricsGetV(metricsGet *GpmMetricsGetType) GpmMetricsGetVType {
 	return GpmMetricsGetVType{metricsGet.convert()}
```
klueska, line 58: Does this need to be updated somehow too? Otherwise, how do the original metrics make their way out of this call?

elezar: The `.convert()` call does also copy the metrics meaning that any metrics in `metricsGet` will be in the result of `convert()`.

klueska: I think we should change this whole thing:
```
// nvml.GpmMetricsGet()
type GpmMetricsGetVType struct {
	metricsGet *nvmlGpmMetricsGetType
}

func (l *library) GpmMetricsGetV(metricsGet *GpmMetricsGetType) GpmMetricsGetVType {
	return GpmMetricsGetVType{metricsGet.convert()}
}
func (metricsGetV GpmMetricsGetVType) V1() Return {
	metricsGetV.metricsGet.Version = 1
	return nvmlGpmMetricsGet(metricsGetV.metricsGet)
}

func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
	metricsGet.Version = GPM_METRICS_GET_VERSION
	return nvmlGpmMetricsGet(metricsGet.convert())
}
```
to this:
```
// nvml.GpmMetricsGet()
type GpmMetricsGetVType struct {
	metricsGet *GpmMetricsGetType
}

func (l *library) GpmMetricsGetV(metricsGet *GpmMetricsGetType) GpmMetricsGetVType {
	return GpmMetricsGetVType{metricsGet}
}
func (metricsGetV GpmMetricsGetVType) V1() Return {
	nvmlMetricsGet := metricsGetV.metricsGet.convert()
	nvmlMetricsGet.Version = 1
	ret := nvmlGpmMetricsGet(&nvmlMetricsGet)
	*metricsGetV.metricsGet = nvmlMetricsGet.convert()
	return ret
}

func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
	nvmlMetricsGet := metricsGet.convert()
	nvmlMetricsGet.Version = GPM_METRICS_GET_VERSION
	ret := nvmlGpmMetricsGet(&nvmlMetricsGet)
	*metricsGetV.metricsGet = nvmlMetricsGet.convert()
	return ret
}
```

pkg/nvml/gpm.go
elezar: Sure. Let's have a look at this at some point in the (near) future.

pkg/dl/dl.go
```diff
@@ -65,6 +65,7 @@ func dlError() error {
 }
 
 func (dl *DynamicLibrary) Open() error {
+	fmt.Printf("%+v\n", dl)
```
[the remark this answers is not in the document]
elezar: Yes. Sorry, was editing in the wrong window here.

pkg/nvml/gpm.go
klueska: I see how your most recent changes will work for `GpmMetricsGet()` (even though I still think we should use an explicit convert) -- but I still don't understand how you believe they work for `GpmMetricsGetV(&metricsGet).V1()` without something like what I have above.

pkg/nvml/gpm.go
```diff
@@ -57,14 +57,22 @@ type GpmMetricsGetVType struct {
...
-	return nvmlGpmMetricsGet(metricsGetV.metricsGet)
+	return nvmlGpmMetricsGetStub(metricsGetV.metricsGet)
 }
 
 func (l *library) GpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
 	metricsGet.Version = GPM_METRICS_GET_VERSION
-	return nvmlGpmMetricsGet(metricsGet.convert())
+	nvmlMetricsGet := metricsGet.convert()
+	ret := nvmlGpmMetricsGetStub(nvmlMetricsGet)
+	metricsGet.Metrics = nvmlMetricsGet.Metrics
```
elezar, line 73: @klueska we create a local instance of an nvml type before the function call and we set the return value explicitly here.

pkg/nvml/gpm.go
elezar: Yes, you're right about `V1()`. I've updated the revision to add a test for that too and also update the implementation.

pkg/nvml/gpm.go
```diff
@@ -51,20 +51,31 @@ func (g *nvmlGpmMetricsGetType) convert() *GpmMetricsGetType {
...
 	metricsGet.Version = GPM_METRICS_GET_VERSION
-	return nvmlGpmMetricsGet(metricsGet.convert())
+	return gpmMetricsGet(metricsGet)
+}
+
+func gpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
+	nvmlMetricsGet := metricsGet.convert()
+	ret := nvmlGpmMetricsGetStub(nvmlMetricsGet)
+	metricsGet.Metrics = nvmlMetricsGet.Metrics
+	return ret
```
elezar, line 78: @klueska following your comments, I could update this whole thing to:
```suggestion
func gpmMetricsGet(metricsGet *GpmMetricsGetType) Return {
	nvmlMetricsGet := metricsGet.convert()
	ret := nvmlGpmMetricsGet(nvmlMetricsGet)
	*metricsGet = *nvmlMetricsGet.convert()
	return ret
```
do you have a preference?

pkg/nvml/gpm.go
klueska: OK, something changed since I made my previous comment. I was still seeing:
```
func (l *library) GpmMetricsGetV(metricsGet *GpmMetricsGetType) GpmMetricsGetVType {
	return GpmMetricsGetVType{metricsGet.convert()}
}
```
whereby we lose our reference to the original `metricsGet` variable as soon as we do `metricsGet.convert()`. The current code seems to basically do what I was suggesting now.

The one difference being that you do:
```
metricsGet.Metrics = nvmlMetricsGet.Metrics
```
instead of:
```
*metricsGet = nvmlMetricsGet.convert()
```
This is equivalent for now -- but your version would need to be updated if any other fields become OUT fields in the future, wheras using `convert()` avoids this.

pkg/nvml/gpm.go
klueska: This is equivalent for now -- but your version would need to be updated if any other fields become OUT fields in the future, wheras using convert() avoids this.

elezar: Ok. Let's switch to the full assignmet to better handle future fields.

klueska (approved):

PR #123, opened 2024-05-16
Bump NVML to 12.4 via 12.1, 12.2, and 12.3

klueska:


pkg/nvml/const.go
```diff
@@ -78,18 +78,6 @@ const (
 	VGPU_NAME_BUFFER_SIZE = 64
 	// GRID_LICENSE_FEATURE_MAX_COUNT as defined in nvml/nvml.h
 	GRID_LICENSE_FEATURE_MAX_COUNT = 3
-	// VGPU_SCHEDULER_POLICY_UNKNOWN as defined in nvml/nvml.h
-	VGPU_SCHEDULER_POLICY_UNKNOWN = 0
```
elezar, line 82: Although this is due to the `nvml.h` change, this is breaking from the point of view of consumers -- especially if we consder that we may be loading and older `libnvidia-ml.so` version that still has these defined.

pkg/nvml/device.go
```diff
@@ -2654,48 +2654,6 @@ func (device nvmlDevice) GetVgpuCapabilities(capability DeviceVgpuCapability) (b
 	return (capResult != 0), ret
 }
 
-// nvml.DeviceGetVgpuSchedulerLog()
-func (l *library) DeviceGetVgpuSchedulerLog(device Device) (VgpuSchedulerLog, Return) {
```
elezar: Do these changes mean that we should do each of the verison bumps as separate PRs? At least we then have the option to branch at the `12.1` update and apply fixes there.

pkg/nvml/zz_generated.api.go
```diff
@@ -184,9 +184,6 @@ var (
 	DeviceGetVgpuCapabilities                       = libnvml.DeviceGetVgpuCapabilities
 	DeviceGetVgpuMetadata                           = libnvml.DeviceGetVgpuMetadata
 	DeviceGetVgpuProcessUtilization                 = libnvml.DeviceGetVgpuProcessUtilization
-	DeviceGetVgpuSchedulerCapabilities              = libnvml.DeviceGetVgpuSchedulerCapabilities
```
elezar, line 187: Since we're generating APIs now, could we call out the removal of these functions in a changelog / release notes? (more of a nice-to-have)

pkg/nvml/const.go
```diff
@@ -78,6 +78,24 @@ const (
 	VGPU_NAME_BUFFER_SIZE = 64
 	// GRID_LICENSE_FEATURE_MAX_COUNT as defined in nvml/nvml.h
 	GRID_LICENSE_FEATURE_MAX_COUNT = 3
+	// VGPU_SCHEDULER_POLICY_UNKNOWN as defined in nvml/nvml.h
+	VGPU_SCHEDULER_POLICY_UNKNOWN = 0
```
elezar, line 84: OK. I see, they re-added them here.

pkg/nvml/nvml.go
```diff
@@ -275,6 +275,15 @@ func nvmlDeviceGetSerial(nvmlDevice nvmlDevice, Serial *byte, Length uint32) Ret
 	return __v
 }
 
+// nvmlDeviceGetModuleId function as declared in nvml/nvml.h
+func nvmlDeviceGetModuleId(nvmlDevice nvmlDevice, ModuleId *uint32) Return {
```
elezar, line 289: Note to self: Can we update `c-for-go` to return `moduleId` here instead of `ModuleId`?

pkg/nvml/nvml.go
```diff
@@ -102,6 +102,25 @@ func nvmlSystemGetProcessName(Pid uint32, Name *byte, Length uint32) Return {
 	return __v
 }
 
+// nvmlSystemGetHicVersion function as declared in nvml/nvml.h
+func nvmlSystemGetHicVersion(HwbcCount *uint32, HwbcEntries *HwbcEntry) Return {
```
elezar, line 106: I don't see a corresponding top-level or interface implementation for this.

pkg/nvml/nvml.go
```diff
@@ -351,6 +351,15 @@ func nvmlDeviceClearCpuAffinity(nvmlDevice nvmlDevice) Return {
 	return __v
 }
 
+// nvmlDeviceGetNumaNodeId function as declared in nvml/nvml.h
+func nvmlDeviceGetNumaNodeId(nvmlDevice nvmlDevice, Node *uint32) Return {
```
elezar, line 355: Question: Does this mean we don't need to use the heuristics that we currently use?

elezar (changes requested):
I think we may be missing some `System*` functions for the `12.3` bump.

A general question is whether we want to separate this into separate PRs to better track the version updates? (We need not release each of these, but we *could*)

pkg/nvml/const.go
klueska: NVML only guarantees backwards compatibility within a major release. Are you saying we want to have a stronger guarantee?

pkg/nvml/device.go
klueska: Either that or just tag the appropriate commits

pkg/nvml/const.go
klueska: Yeah, they were removed in 12.1 for some reason and then readded in 12.2

pkg/nvml/nvml.go
klueska: I'm not sure. We could / should look into it.

pkg/nvml/nvml.go
klueska: I would say so. But we need to be aware that this is only available in later releases.

pkg/nvml/nvml.go
klueska: Its there already, actually:
https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/system.go#L53

In each of these versions bumps a whole bunch of functions were moved around, so there are a bunch of `+`s and `-`s on functions that are really just moves within the file.

pkg/nvml/zz_generated.api.go
klueska: They were added back in. There were however, 2 legitimate removals:
```
-   CcuGetStreamState() (int, Return)
-   CcuSetStreamState(int) Return
```
Not sure if these were intentional...

```

pkg/nvml/device.go
elezar: I suppose they're still available even after a merge.

pkg/nvml/nvml.go
elezar: Yes, we would have to support a fallback.

klueska:
Do you have further comments here? Or more comments to add to my responses?

elezar (approved):

elezar:
> Do you have further comments here? Or more comments to add to my responses?

No. Looks good. Thanks for the responses.

klueska:
Let's fix the GPM metrics before merging this so that the change appears in all of the versions 12.0-12.4

PR #125, opened 2024-06-19
Return golang errors instead of nvml.Return values

elezar:
These changes move from using `nvml.Return` values to using native Golang `error` values, making code more idiomatic.

Note that this requires that `error.Is` is used instead of checking for equality in the returned value.

For example:
```
current, pending, err := device.GetMigMode()
if errors.Is(err, nvml.ERROR_NOT_SUPPORTED) {
	fmt.Printf("MIG is not supported for device %v\n", i)

} else if err != nil {
	log.Fatalf("Error getting MIG mode for device at index %d: %v", i, err)
}
fmt.Printf("MIG mode for device %v: current=%v pending=%v\n", i, current, pending)
```

would become:
```
current, pending, err := device.GetMigMode()
if errors.Is(err, nvml.ERROR_NOT_SUPPORTED) {
	fmt.Printf("MIG is not supported for device %v\n", i)

} else if err != nil {
	log.Fatalf("Error getting MIG mode for device at index %d: %v", i, err)
}
fmt.Printf("MIG mode for device %v: current=%v pending=%v\n", i, current, pending)
```


klueska:
Your two example's in the description are the same .. I think you wanted one of them to check euqility instead of using `errors.Is()`.

That said -- what will happen if someone does do `err == nvml.ERROR_NOT_SUPPORTED`? Will it be `false` even when `errors.Is(err, nvml.ERROR_NOT_SUPPORTED)` would be `true`?

pkg/nvml/nvml.h
klueska, line 1: None of the changes in this file look intentional...

pkg/nvml/return.go
```diff
@@ -20,17 +20,76 @@ import (
 
 // nvml.ErrorString()
 func (l *library) ErrorString(r Return) string {
-	return r.Error()
+	return r.String()
 }
```
klueska, line 24: How does an end-user ever get ahold of a `Return` value to pass it to this function?

PR #129, opened 2024-08-30
Add VgpuTypeIDToUint32 helper function

PiotrProkop:
This PR adds new `VgpuTypeIdToUint32` helper function that allows to retrieve underlying `uint32` representation of `nvmlVgpuTypeId`:

```go
type nvmlVgpuTypeId uint32
```

this makes it easier to programmatically create vgpus , for example:

```sh
echo $(uuidgen) > /sys/bus/pci/devices/0000\:06\:00.0/mdev_supported_types/nvidia-<typeID>/create
```

where `<typeID>` is this underlying `uint32`.

In previous versions of `go-nvml` I could just do type conversion to `uint32` directly as `device.GetSupportedVgpus()` was returning:
```
type VgpuTypeId uint32
```

instead of Interface.


pkg/nvml/cgo_helpers_static.go
```diff
@@ -73,3 +73,7 @@ func unpackPCharString(str string) (*C.char, *struct{}) {
 	h := (*stringHeader)(unsafe.Pointer(&str))
 	return (*C.char)(h.Data), cgoAllocsUnknown
 }
+
+func VgpuTypeIdToUint32(typeId VgpuTypeId) uint32 {
```
elezar, line 77: I know that there are exceptions, but in general we try to expose NVML-functionality in these bindings and this goes against that a bit.

Does your example not indicate the need for a `String()` function off interface instead of an explicit `*ToUint32` function?

elezar (commented):
Thanks @PiotrProkop.

It would be good to get @klueska to also take a look at this.

pkg/nvml/cgo_helpers_static.go
PiotrProkop: I got impression that interface for VgpuTypeId is auto generated based on NVML lib funcs, so I wanted to propose less intrusive option. `String` or `uint32` both works for me. I wanted to show that newer `go-nvml` version is missing some info that was present in older version.

elezar: I have looked at this a bit. What about doing something like we do for the `Interface` type where we have the function:
```
Extensions() ExtendedInterface
```

We could to the same for `VgpuTypeId`:
```
//go:generate moq -rm -out mock/extendedvgputypeid.go -pkg mock . ExtendedVgpuTypeId:ExtendedVgpuTypeId
type ExtendedVgpuTypeId interface {
       ID() uint32
}
```

I've pushed a commit with a proposal on top. Let me know what you think.

pkg/nvml/zz_generated.api.go
```diff
@@ -668,6 +667,7 @@ type Interface interface {
 	VgpuTypeGetName(VgpuTypeId) (string, Return)
 	VgpuTypeGetNumDisplayHeads(VgpuTypeId) (int, Return)
 	VgpuTypeGetResolution(VgpuTypeId, int) (uint32, uint32, Return)
+	Extensions() ExtendedInterface
```
klueska, line 670: Why / how did this get generated at the bottom? The generation code sorts the functions alphabetically ...

pkg/nvml/vgpu_extensions.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nvml
+
+func (vgpuTypeId *nvmlVgpuTypeId) Extensions() ExtendedVgpuTypeId {
```
PiotrProkop, line 17: I think `vgpuTypeId` doesn't have to be a pointer receiver.

pkg/nvml/vgpu_extensions.go
```diff
@@ -0,0 +1,27 @@
...
+	return vgpuTypeId
+}
+
+// ID returns the numeric representaion of the vgpuTypeId.
+func (vgpuTypeId *nvmlVgpuTypeId) ID() uint32 {
```
PiotrProkop, line 22: I think `vgpuTypeId` doesn't have to be a pointer receiver.

pkg/nvml/cgo_helpers_static.go
PiotrProkop: Looks good to me! Thanks for looking into this.

pkg/nvml/zz_generated.api.go
rajatchopra: see the code diff in genrateapi.go

pkg/nvml/vgpu_extensions.go
rajatchopra: Do you plan to fix this @PiotrProkop ?
Remove pointer, that is.

rajatchopra (commented):
Not sure if we really need this helper extension.

rajatchopra:
@PiotrProkop Can you revive this PR? If not paid attention to, we may want to close out older PRs that we have lived well without.

klueska:
I think this would supercede it: https://github.com/NVIDIA/go-nvml/pull/178

PiotrProkop:
@elezar took over my PR, so I'll defer to him.

cdesiniotis:
Superseded by https://github.com/NVIDIA/go-nvml/pull/178. Closing.

PR #130, opened 2024-09-19
Fix the bug that some functions do not populate the Version field, resulting in failure of the call.

xxsoul:
Fix the bug that the Version field of ProcessDetailList is not assigned a value(must be nvmlProcessDetailList_v1) when calling the underlying nvmlDeviceGetRunningProcessDetailList function.

If the value is not populated (default is 0), the function will return nvmlReturn_t=25(Argument version mismatch)

### DIFF
--- a/pkg/nvml/nvml.go
+++ b/pkg/nvml/nvml.go
@@ -3100,6 +3100,7 @@ func nvmlDeviceSetPowerManagementLimit_v2(nvmlDevice nvmlDevice, PowerValue *Pow
 // nvmlDeviceGetSramEccErrorStatus function as declared in nvml/nvml.h
 func nvmlDeviceGetSramEccErrorStatus(nvmlDevice nvmlDevice, Status *EccSramErrorStatus) Return {
 	cnvmlDevice, _ := *(*C.nvmlDevice_t)(unsafe.Pointer(&nvmlDevice)), cgoAllocsUnknown
+	Status.Version = C.nvmlEccSramErrorStatus_v1
 	cStatus, _ := (*C.nvmlEccSramErrorStatus_t)(unsafe.Pointer(Status)), cgoAllocsUnknown
 	__ret := C.nvmlDeviceGetSramEccErrorStatus(cnvmlDevice, cStatus)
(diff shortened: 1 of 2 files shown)


pkg/nvml/nvml.go
```diff
@@ -1224,6 +1224,7 @@ func nvmlDeviceGetMPSComputeRunningProcesses_v3(nvmlDevice nvmlDevice, InfoCount
 // nvmlDeviceGetRunningProcessDetailList function as declared in nvml/nvml.h
 func nvmlDeviceGetRunningProcessDetailList(nvmlDevice nvmlDevice, Plist *ProcessDetailList) Return {
 	cnvmlDevice, _ := *(*C.nvmlDevice_t)(unsafe.Pointer(&nvmlDevice)), cgoAllocsUnknown
+	Plist.Version = C.nvmlProcessDetailList_v1
```
klueska, line 1227: This should not be set here. This is generated code. It needs to be set in the wrapper passed into this function.

xxsoul: Is it possible to add logic to set the version in file "device.go"? I noticed from nvml.h that nvmlProcessDetailList_v1 is actually calculated by the macro definition (size_of(nvmlProcessDetailList_v1_t) | (1 << 24)), and it looks like we need to define some constants to hold such kind of values.

klueska: I imagine it can be done the same as here: https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/device.go#L826

xxsoul: Ingenious way, I reworked my submission as in [code](https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/device.go#L826) :-)

elezar:
@xxsoul could you please sign-off your contributions as required by the DCO?

elezar:
Alternatively I will re-implement this in a follow-up.

elezar:
@xxsoul I created #139 to address this.

cdesiniotis:
Closing as this should be addressed in #139

PR #131, opened 2024-10-10
Add -z,lazy to LDFLAGS

elezar:
This fixes undefined symbol errors on platforms where -z,lazy may not be the default.


klueska (approved):

DanielGibson:
Probably the safest fix would be to use `dlsym()` to load the libnvidia-ml.so functions, instead of kinda-linking-against-it-but-not-really and relying on lazy binding

elezar:
As @klueska points out. This is not the right place to do this. It needs to be added explicitly in the application being built.

Issue #134, opened 2024-11-07
The device.GetSupportedGraphicsClocks only return one clock value


wangchen615:
https://github.com/NVIDIA/go-nvml/blob/852755d4e0bdbd08328e455217e28ac2f93e4a25/pkg/nvml/device.go#L575

Is this the expected behavior? But when I am doing nvidia-smi to get the supported clock values, there are a list:

`$ ~# nvidia-smi -q -d SUPPORTED_CLOCKS

==============NVSMI LOG==============

Timestamp                                 : Thu Nov  7 02:51:00 2024
Driver Version                            : 535.183.01
CUDA Version                              : 12.2

Attached GPUs                             : 2
GPU 00000000:04:01.0
    Supported Clocks
        Memory                            : 877 MHz
            Graphics                      : 1380 MHz
            Graphics                      : 1372 MHz
            Graphics                      : 1365 MHz
            Graphics                      : 1357 MHz
            Graphics                      : 1350 MHz
            Graphics                      : 1342 MHz
            Graphics                      : 1335 MHz
            Graphics                      : 1327 MHz
            Graphics                      : 1320 MHz
            Graphics                      : 1312 MHz
            Graphics                      : 1305 MHz
            Graphics                      : 1297 MHz
            Graphics                      : 1290 MHz
            Graphics                      : 1282 MHz
            Graphics                      : 1275 MHz
            Graphics                      : 1267 MHz
            Graphics                      : 1260 MHz
            Graphics                      : 1252 MHz
            Graphics                      : 1245 MHz
            Graphics                      : 1237 MHz
            Graphics                      : 1230 MHz
            Graphics                      : 1222 MHz
            Graphics                      : 1215 MHz
            Graphics                      : 1207 MHz
            Graphics                      : 1200 MHz
            Graphics                      : 1192 MHz
            Graphics                      : 1185 MHz
            Graphics                      : 1177 MHz
            Graphics                      : 1170 MHz
            Graphics                      : 1162 MHz
            Graphics                      : 1155 MHz
            Graphics                      : 1147 MHz
            Graphics                      : 1140 MHz
            Graphics                      : 1132 MHz
            Graphics                      : 1125 MHz
            Graphics                      : 1117 MHz
            Graphics                      : 1110 MHz
            Graphics                      : 1102 MHz
            Graphics                      : 1095 MHz
            Graphics                      : 1087 MHz
            Graphics                      : 1080 MHz
            Graphics                      : 1072 MHz
[...]

PR #135, opened 2024-11-07
feat: probe all supported mem and graphics frequencies

rootfs:
Fix issue #134 

There is a hack in this fix, the generated api is manually modified, since I don't know how it is generated in the first place.


rajatchopra (changes requested):
Changing the return type of existing API would mean a breaking change. Cannot do so unless we introduce a version on the API with a backward compatible translator.

Although the existing API has a misleading name, we may have to live with the confusing option of creating another API (with another name) that does the right thing. e.g. DeviceGetAllSupportedMemoryClocks()

klueska:
Actually, for this one, I would call it a bug and fix it. Just returning a single uint32 is meaningless for the purpose of the original C API, and was implemented incorrectly when reflected into the go API. It should always have been returning a list.

rajathagasthya:
This branch can't be merged as-is and doesn't allow maintainer edits, so closing it in favor of a fresh PR. The new PR will fix the signature and generate mocks appropriately.

Issue #128, opened 2024-08-23
`GetConfComputeGpuAttestationReport` does not accept a nonce


derpsteb:
Hey,

the function `GetConfComputeGpuAttestationReport` does currently not accept a nonce for the report generation. This is insecure for CC use cases because an attacker might record a valid report once and replay it to a victim multiple times/on different machines, potentially with a modified GPU attached. 

We have implemented this in our fork [here](https://github.com/edgelesssys/go-nvml/commit/46c94d0840e405a33adef735ab1671ee8ec0229f). This works fine so it seems to be supported by the runtime. 

Could you update that function?
Thanks,
Otto

klueska:
I checked with the NVML team and they confirmed that if you pass a nonce in the calling struct to the underlying C API, then it will be used by the driver (rather than generating one on the fly).

I'll need to think about the right API for this because we don't like to stray too far from the underlying C API (with the exception of transforming inputs meant as return values to actual return values). In this case though, the argument to the C API is both a valid input AND a place to store the return value...

derpsteb:
Hey. Did you have a chance to consider this? Would be cool if we could drop our fork.

jmomort:
Note that the python library NVIDIA created does indeed use the nonce. https://github.com/NVIDIA/nvtrust/blob/02b2d22211fb506ab5dce0a23b15858d993cf6d0/guest_tools/gpu_verifiers/local_gpu_verifier/src/verifier/nvml/__init__.py#L195-L199

jmomort:
Resolved with https://github.com/NVIDIA/go-nvml/pull/140

klueska:
I'm looking at introducing an option to generically pass the return-value-as-argument (similar to what the C API does) in order to future proof support for this (and any future changes like it).

PR #140, opened 2025-01-13
Allow to pass nonce to attestation report 

yoihito:
We found out that currently there is no way to pass a nonce to `DeviceGetConfComputeGpuAttestationReport`. We need to pass nonce for security reasons. We have added a new method `DeviceGetConfComputeGpuAttestationReportWithNonce` to keep the old one `DeviceGetConfComputeGpuAttestationReport` backwards compatible.


jmomort:
@klueska can you review? This should resolve https://github.com/NVIDIA/go-nvml/issues/128

klueska:
Was this closed by accident? As far as I'm aware, we haven't added this feature yet.

PR #142, opened 2025-04-21
migrate to golangci-lint v2

tariq1890:
The following lint issues have been fixed

```
                      ^
pkg/nvml/mock/dgxa100/dgxa100.go:22:1: File is not properly formatted (goimports)

^
pkg/nvml/mock/dgxa100/dgxa100.go:111:23: composites: github.com/NVIDIA/go-nvml/pkg/nvml.Memory struct literal uses unkeyed fields (govet)
                MemoryInfo:         nvml.Memory{42949672960, 0, 0},
                                    ^
gen/nvml/generateapi.go:106:2: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
        fmt.Fprintf(writer, header)
        ^
gen/nvml/generateapi.go:115:4: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
                        fmt.Fprintf(writer, comment)
                        ^
gen/nvml/generateapi.go:130:3: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
                fmt.Fprintf(writer, comment)
                ^
gen/nvml/generateapi.go:273:15: ST1005: error strings should not end with punctuation or newlines (staticcheck)
                return nil, fmt.Errorf("walking %s: %w\n", sourceDir, err)
                            ^
pkg/nvml/device.go:2194:12: QF1011: could omit type uint32 from declaration; it will be inferred from the right-hand side (staticcheck)
        var count uint32 = info.InstanceCount
                  ^
pkg/nvml/device.go:2305:12: QF1011: could omit type uint32 from declaration; it will be inferred from the right-hand side (staticcheck)
        var count uint32 = info.InstanceCount
                  ^
pkg/nvml/gpm.go:33:2: S1001: should copy arrays using assignment instead of using a loop (staticcheck)
        for i := range g.Metrics {
        ^
pkg/nvml/gpm.go:46:2: S1001: should copy arrays using assignment instead of using a loop (staticcheck)
        for i := range g.Metrics {
        ^
                   ^

```

### DIFF
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -42,7 +42,7 @@ jobs:
       with:
         go-version: ${{ env.GOLANG_VERSION }}
     - name: Lint
-      uses: golangci/golangci-lint-action@v6
+      uses: golangci/golangci-lint-action@v7
       with:
         version: latest
         args: -v --timeout 5m
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,19 +1,34 @@
+version: "2"
 linters:
-  enable:
-    - goimports
   disable:
     - asciicheck
     - contextcheck
+    - errcheck
     - forcetypeassert
     - gocritic
     - godot
-    - gofmt
     - misspell
-    - stylecheck
-
-run:
-  deadline: 10m
-
-linters-settings:
-  goimports:
-    local-prefixes: github.com/NVIDIA/go-nvml
+  exclusions:
+    generated: lax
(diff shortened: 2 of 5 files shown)


yxxhero (approved):

ArangoGutierrez (approved):
Thanks for this Tariq, v2 is very picky

elezar (approved):

PR #143, opened 2025-04-21
sync vendor/ dir with go mod

tariq1890:
These issues have been uncaught as the `check` job has been disabled. https://github.com/NVIDIA/go-nvml/pull/104

Once the recent PRs are merged, we should be in good shape to turn it back on


ArangoGutierrez (approved):

elezar:
@tariq1890 since go-nvml is consumed as a library we do not vendor dependencies. I think we can update the check target to not include the vendor folder.

PR #144, opened 2025-04-30
[CI] re-enable the check stage

tariq1890:
The following lint errors have been fixed as part of this PR changeset 
```
pkg/nvml/device.go:71:26: func GpuInstanceInfo.convert is unused (unused)
func (g GpuInstanceInfo) convert() nvmlGpuInstanceInfo {
                         ^
pkg/nvml/device.go:100:30: func ComputeInstanceInfo.convert is unused (unused)
func (c ComputeInstanceInfo) convert() nvmlComputeInstanceInfo {
                             ^
pkg/nvml/event_set.go:26:20: func EventData.convert is unused (unused)
func (e EventData) convert() nvmlEventData {

gen/nvml/generateapi.go:130:3: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
  		fmt.Fprintf(writer, comment)
  		^
gen/nvml/generateapi.go:137:3: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
  		fmt.Fprintf(writer, output)
  	
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ EXAMPLE_TARGETS := $(patsubst %,example-%, $(EXAMPLES))
 CMDS := $(patsubst ./cmd/%/,%,$(sort $(dir $(wildcard ./cmd/*/))))
 CMD_TARGETS := $(patsubst %,cmd-%, $(CMDS))
 
-CHECK_TARGETS := golangci-lint
+CHECK_TARGETS := validate-modules golangci-lint
 
 MAKE_TARGETS := binary build all fmt generate test coverage check examples update-nvml-h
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -72,6 +72,12 @@ coverage: test
 	cat $(COVERAGE_FILE) | grep -v "_mock.go" > $(COVERAGE_FILE).no-mocks
 	go tool cover -func=$(COVERAGE_FILE).no-mocks
 
+validate-modules:
+	@echo "- Verifying that the dependencies have expected content..."
+	go mod verify
+	@echo "- Checking for any unused/missing packages in go.mod..."
+	go mod tidy
```
elezar, line 79: Note that we would need to run a diff or something here to ensure no differences. 
I think something like the following should work (since we're not using vendoring):
```
git diff --exit-code HEAD -- go.mod go.sum
```

tariq1890: True, thanks for catching that

tariq1890: Done, please check now

ArangoGutierrez (approved):
Thanks for this Tariq

elezar (approved):

PR #155, opened 2025-07-17
Add script to generate a raw cgo-level interface for go-nvml

klueska:
This interface can be used in conjunction with the standard "wrapper"
    interface, including the use its interface abstractions like Devices,
    GPUInstances, ComputeInstances, etc.
    
    It is invoked as nvml.CgoAPI.<api-call>, following the original function
    signatures of the C-API, where there is only a single return value and
    additional return values are passed as arguments.
    
    An example usage is:
    ```
            // Start with using the CgoAPI
            ret := nvml.CgoAPI.Init()
            defer nvml.Shutdown()
            require.Equal(t, SUCCESS, ret, "Init should succeed")
    
            var count uint32
            ret = nvml.CgoAPI.DeviceGetCount(&count)
            require.Equal(t, SUCCESS, ret, "DeviceGetCount should succeed")
    
            var device Device
            ret = nvml.CgoAPI.DeviceGetHandleByIndex(0, &device)
            require.Equal(t, SUCCESS, ret, "DeviceGetHandleByIndex should succeed")
    
            // Now call Device interface methods directly
            name, ret := device.GetName()
            require.Equal(t, SUCCESS, ret, "Device.GetName should succeed")
            t.Logf("Device.GetName: %s", name)
    
            uuid, ret := device.GetUUID()
            require.Equal(t, SUCCESS, ret, "Device.GetUUID should succeed")
            t.Logf("Device.GetUUID: %s", uuid)
    
            brand, ret := device.GetBrand()
            require.Equal(t, SUCCESS, ret, "Device.GetBrand should succeed")
            t.Logf("Device.GetBrand: %v", brand)
    
            busType, ret := device.GetBusType()
            require.Equal(t, SUCCESS, ret, "Device.GetBusType should succeed")
            t.Logf("Device.GetBusType: %v", busType)
    ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,9 +151,14 @@ bindings: .create-bindings .strip-autogen-comment .strip-nvml-h-linenumber
 		go fmt types_gen.go; \
 	cd -> /dev/null
 	rm -rf $(PKG_BINDINGS_DIR)/nvml.yml $(PKG_BINDINGS_DIR)/cgo_helpers.go $(PKG_BINDINGS_DIR)/types.go $(PKG_BINDINGS_DIR)/_obj
-	go run $(GEN_BINDINGS_DIR)/generateapi.go \
+	go run -tags generateapi \
+		$(GEN_BINDINGS_DIR)/generateapi.go \
 		--sourceDir $(PKG_BINDINGS_DIR) \
 		--output $(PKG_BINDINGS_DIR)/zz_generated.api.go
+	go run -tags generatecgoapi \
+		$(GEN_BINDINGS_DIR)/generatecgoapi.go \
+		--sourceDir $(PKG_BINDINGS_DIR) \
+		--output $(PKG_BINDINGS_DIR)/zz_generated.cgoapi.go
 	make fmt
 
 .strip-autogen-comment: SED_SEARCH_STRING := // WARNING: This file has automatically been generated on
--- a/gen/nvml/generateapi.go
+++ b/gen/nvml/generateapi.go
@@ -14,6 +14,7 @@
 # limitations under the License.
 **/
 
+//go:build generateapi
 package main
 
 import (
--- a/gen/nvml/generateapi.go
+++ b/gen/nvml/generateapi.go
@@ -29,6 +29,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"text/template"
 	"unicode"
 )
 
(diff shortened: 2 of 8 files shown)


gen/nvml/generateapi.go
```diff
@@ -418,3 +433,76 @@ func isPublic(name string) bool {
...
+	}
+
+	return builder.String(), nil
+}
+
+func generateHandleHelper(input GeneratableInterfacePoperties) (string, error) {
+	var builder strings.Builder
+
+	// Function comment
+	builder.WriteString(fmt.Sprintf("// %sHandle attempts to convert a %s to an %s.\n",
```
elezar, line 460: This may be easier to read if we use go templates instead.

gen/nvml/generatecgoapi.go
```diff
@@ -0,0 +1,693 @@
...
+	sb.WriteString(structDef)
+	for _, m := range methods {
+		sb.WriteString(m.Code)
+	}
+
+	return sb.String(), nil
+}
+
+// generateHeader returns the file header
+func generateHeader() string {
```
elezar, line 131: Could we share the `generateHeader` functions between here and `generateapi`?

Makefile
```diff
@@ -151,9 +151,14 @@ bindings: .create-bindings .strip-autogen-comment .strip-nvml-h-linenumber
 		go fmt types_gen.go; \
 	cd -> /dev/null
 	rm -rf $(PKG_BINDINGS_DIR)/nvml.yml $(PKG_BINDINGS_DIR)/cgo_helpers.go $(PKG_BINDINGS_DIR)/types.go $(PKG_BINDINGS_DIR)/_obj
-	go run $(GEN_BINDINGS_DIR)/generateapi.go \
+	go run -tags generateapi \
```
elezar, line 154: OK, so we're using build tags to select between the two different `main` packages. Does putting these in their own folders instead simplify things?

klueska: I guess we could, but I really want to keep each of them as their own standalone scripts without an attempt to share code between them (meaning each would likely be a folder with a single file each in them).

gen/nvml/generateapi.go
klueska: Yes, definitely. Updated.

klueska: I just noticed your comment is in the original generateapi.go file (which is not part of the PR). I added the template stuff only to the new script being added as part of this PR. We can revisit the other one in a separate PR.

pkg/nvml/cgoapi_test.go
```diff
@@ -0,0 +1,539 @@
...
+
+	var attestationReport4 ConfComputeGpuAttestationReport
+	attestationReport4.Nonce = testNonce
+
+	// Verify nonce was set correctly
+	for i := 0; i < 32; i++ {
+		require.Equal(t, testNonce[i], attestationReport4.Nonce[i], "Nonce field should be set correctly")
+	}
+
+	ret = CgoAPI.DeviceGetConfComputeGpuAttestationReport(device, &attestationReport4)
```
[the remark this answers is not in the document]
steven-bellock: If this is being tested with a real GPU then that is expected. The GPU will generate a random value for its nonce in the `MEASUREMENTS` response.
image

gen/nvml/generatecgoapi.go
elezar: No. It's fine.

klueska:
@elezar what are your thoughts on the name of "CgoAPI"? I'm not sold on it, but couldn't come up with anything else.

elezar (commented):
My one question is whether we want to use the `cgoapi` internally to provide more idiomatic go wrappers? Maybe as part of something like #125?

(This would obviously be a follow-up).

pkg/nvml/lib.go
```diff
@@ -144,7 +144,7 @@ func (l *library) close() (rerr error) {
 	return nil
 }
 
-// Default all versioned APIs to v1 (to infer the types)
+// BEGIN_UNVERSIONED_FUNCTIONS
```
elezar, line 147: It is common for machine readable comments not have a leading spaces. What about:

```suggestion
//generatecgoapi: BEGIN_UNVERSIONED_FUNCTIONS
```
(and below)

pkg/nvml/cgoapi_test.go
```diff
@@ -0,0 +1,539 @@
...
+		var mode ComputeMode
+		ret = CgoAPI.DeviceGetComputeMode(device, &mode)
+		require.Equal(t, SUCCESS, ret, "DeviceGetComputeMode should succeed")
+	} else {
+		t.Logf("No devices available for testing, skipping device-specific tests")
+	}
+}
+
+// TestCgoAPIDeviceMethods exercises various device-related CgoAPI methods.
+func TestCgoAPIDeviceMethods(t *testing.T) {
```
elezar, line 65: It's not quite clear to me how this is different from the functionality tested in the previous test case -- with the exception of non-overlapping function calls.

rajathagasthya:
@klueska Should we still plan to do this? It's been in draft for a long time now.

klueska:
In principle, I think it's probably still a good idea. That said, I haven't looked at the specifics of this PR in a long time, so its would probably need to be properly revisited.

PR #159, opened 2025-08-22
Update DeviceGetconfComputeGpuAttestationReport to take struct as input

klueska:
This is a BREAKING CHANGE, but one that is necessary, because the API is unusable in its current form. In the C API, the input struct us used to pass _both_ input _and_ output. In our original implementation we didn't take this struct as input, so there was no way to get this input passed into the function. The API has now been updated to correct this.

One option could have been to just add the single input field as an input to the API call on its own, but this is not scalable in cases where one my add more input fields to the struct in the future.

In the end, (even though this goes against our other wrapper API calls), we decided it was best to just pass the whole struct and assume its usage as both input and output just like the C API.

We plan to revisit how to more comprehensively accommodate APIS such as this in the future.


elezar (approved):
LGTM.

PR #165, opened 2025-10-17
Fix DeviceGetRunningProcessDetailList API

chenhengqi:
The DeviceGetRunningProcessDetailList API constantly returns ERROR_INSUFFICIENT_SIZE since we did not setup ProcessDetailList properly. Let's allocate memory for ProcessDetailList::ProcArray before invoking nvmlDeviceGetRunningProcessDetailList() so that we can retrieve the process list. Note that the memory allocation has to be done in cgo otherwise we will hit a Golang runtime error.

### DIFF
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -2931,10 +2931,56 @@ func (l *library) DeviceGetRunningProcessDetailList(device Device) (ProcessDetai
 }
 
 func (device nvmlDevice) GetRunningProcessDetailList() (ProcessDetailList, Return) {
+	return deviceGetRunningProcessDetailList(device)
+}
+
+func deviceGetRunningProcessDetailList(device nvmlDevice) (ProcessDetailList, Return) {
 	var plist ProcessDetailList
 	plist.Version = STRUCT_VERSION(plist, 1)
-	ret := nvmlDeviceGetRunningProcessDetailList(device, &plist)
-	return plist, ret
+	plist.NumProcArrayEntries = 1
+
+	for {
+		// Allocate memory in cgo for ProcessDetailList::ProcArray
+		// We can't simply use a unsafe.Pointer of Go slice here
+		// otherwise it will trigger the following error:
+		//   runtime error: cgo argument has Go pointer to unpinned Go pointer
+		count := plist.NumProcArrayEntries
+		cptr := malloc(uintptr(count) * unsafe.Sizeof(ProcessDetail_v1{}))
+		if cptr == nil {
+			return plist, ERROR_MEMORY
+		}
+
+		plist.ProcArray = (*ProcessDetail_v1)(cptr)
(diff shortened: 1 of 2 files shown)


pkg/nvml/device.go
```diff
@@ -2931,10 +2931,56 @@ func (l *library) DeviceGetRunningProcessDetailList(device Device) (ProcessDetai
...
+	return deviceGetRunningProcessDetailList(device)
+}
+
+func deviceGetRunningProcessDetailList(device nvmlDevice) (ProcessDetailList, Return) {
 	var plist ProcessDetailList
 	plist.Version = STRUCT_VERSION(plist, 1)
-	ret := nvmlDeviceGetRunningProcessDetailList(device, &plist)
-	return plist, ret
+	plist.NumProcArrayEntries = 1
+
```
klueska: Why is it not sufficient to just do this:
```
var plist ProcessDetailList
	plist.Version = STRUCT_VERSION(plist, 1)
	plist.NumProcArrayEntries = 1
	plist.ProcArray = &(make([]ProcessDetail_v1, plist.NumProcArrayEntries)[0])
	...
```

Does NVML realloc the space that `plist.ProcArray` points to inside the call?

klueska: Actually, I just looked it up -- it says this:
> To determine the size of the plist->procArray array to allocate, call the function with plist->numProcArrayEntries set to zero and plist->procArray set to NULL. The return code will be either NVML_ERROR_INSUFFICIENT_SIZE (if there are valid processes of type plist->mode to report on, in which case the plist->numProcArrayEntries field will indicate the required number of entries in the array) or NVML_SUCCESS (if no processes of type plist->mode exist).

Which means we should be following the same pattern as e.g.:
https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/device.go#L993

klueska: With no need to introduce all of this c-level malloc wrapper, etc.

chenhengqi: Have you run it locally? I did encounter a runtime error.

chenhengqi:
A gentle ping...

@klueska @elezar

rajathagasthya:
@chenhengqi @klueska Is this still relevant? I don't have sufficient context on this, so appreciate any help here.

pkg/nvml/device.go
rajatchopra: Agree. Here. Tried to reproduce this and I get a runtime error:
```
panic: runtime error: cgo argument has Go pointer to unpinned Go pointer
```

rajatchopra (approved):
Checked the branch and tested it manually. It works.
```
return code for device process list GPU-16a42815-e60a-0bb9-0614-566150ee6848: Success
	process list: {16777240 0 1 0xc00001e3e0}
```

Thanks @chenhengqi 

/lgtm

PR #173, opened 2026-03-03
add workflow to mark issues as stale in case of no activity

rahulait:
Adds a workflow which runs periodically and marks issues as stale which has no activity for pre-defined period.


JunAr7112:
LGTM @rahulait did you test the Github action already?

rahulait:
Yeah, its the same we have in other repos, just copy paste to all repos

JunAr7112 (approved):

PR #176, opened 2026-04-10
Fix GPM query device support version

gjulianm:
Use the raw GPM support version expected by NVML so the example no longer fails with argument version mismatch, and add regression coverage for both helper paths. 

Should fix https://github.com/NVIDIA/go-nvml/issues/168


cdesiniotis (approved):
LGTM. Thanks @gjulianm for the contribution.

PR #178, opened 2026-05-05
vgpu: expose underlying uint32 via VgpuTypeId.GetID()

xpahos:
While iterating supported/creatable vGPU types there is no way to map back to the legacy mediated-device profile names (e.g. "nvidia-577") exposed under /sys/.../nvidia/creatable_vgpu_types — those entries are keyed by the raw nvmlVgpuTypeId uint32, and GetName() returns the device-specific name instead.
Add GetID() uint32 on nvmlVgpuTypeId so callers can read the underlying handle.

### DIFF
--- a/pkg/nvml/vgpu.go
+++ b/pkg/nvml/vgpu.go
@@ -42,6 +42,19 @@ func (vgpuTypeId nvmlVgpuTypeId) GetClass() (string, Return) {
 	return string(vgpuTypeClass[:clen(vgpuTypeClass)]), ret
 }
 
+// nvml.VgpuTypeGetID()
+// It doesn't have an NVML C API symbol because the base type `nvmlVgpuTypeId`
+// is a non-exported type defined as `uint32`. When using the `VgpuTypeId` type,
+// it is not possible to read the underlying value without using reflection.
+// This method adds an idiomatic Go getter to access the type's value.
+func (l *library) VgpuTypeGetID(vgpuTypeId VgpuTypeId) uint32 {
+	return vgpuTypeId.GetID()
+}
+
+func (vgpuTypeId nvmlVgpuTypeId) GetID() uint32 {
+	return uint32(vgpuTypeId)
+}
+
 // nvml.VgpuTypeGetName()
 func (l *library) VgpuTypeGetName(vgpuTypeId VgpuTypeId) (string, Return) {
 	return vgpuTypeId.GetName()
--- a/pkg/nvml/zz_generated.api.go
+++ b/pkg/nvml/zz_generated.api.go
@@ -404,6 +404,7 @@ var (
 	VgpuTypeGetFrameRateLimit                        = libnvml.VgpuTypeGetFrameRateLimit
 	VgpuTypeGetFramebufferSize                       = libnvml.VgpuTypeGetFramebufferSize
 	VgpuTypeGetGpuInstanceProfileId                  = libnvml.VgpuTypeGetGpuInstanceProfileId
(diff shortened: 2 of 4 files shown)


pkg/nvml/zz_generated.api.go
klueska, line 1: There should not be such a large diff here

pkg/nvml/vgpu.go
```diff
@@ -42,6 +42,10 @@ func (vgpuTypeId nvmlVgpuTypeId) GetClass() (string, Return) {
 	return string(vgpuTypeClass[:clen(vgpuTypeClass)]), ret
 }
 
+func (vgpuTypeId nvmlVgpuTypeId) GetID() uint32 {
+	return uint32(vgpuTypeId)
+}
+
```
klueska, line 57: Normally we *only* include API calls in this library that are also part of the C based API. However, since we do this translation of the `VgpuTypeId` to an interface, there is no good way to get the original value back out of it without a helper method like this, so I think it's probably OK to make an exception.

That said, you would also need an equivalent method hanging off of the `library` type called VgpuTypeGetID().

xpahos: Added VgpuTypeGetID. Without these functions, unfortunately, the only way to retrieve the value would be through reflection, which I’d rather not use.

pkg/nvml/zz_generated.api.go
xpahos: The generator ./pkg/nvml/zz_generated.api.go does not automatically run `go fmt`, fixed it manually.

xpahos:
@klueska sorry to bother you, do I need to make any other changes?

xpahos:
@klueska gentle ping. Is there anything else needed from my side for this PR?

xpahos:
@klueska sorry for the extra ping — just checking whether there’s anything else I should change here when you have a chance.

xpahos:
@cdesiniotis @tariq1890 sorry to bother you, but I just noticed you’ve merged the latest PRs. Could you tell me who I should contact to have these changes merged, or perhaps you could merge them yourself?

tariq1890:
@xpahos Can you rebase this PR?

xpahos:
> @xpahos Can you rebase this PR?

done

pkg/nvml/vgpu.go
```diff
@@ -42,6 +42,15 @@ func (vgpuTypeId nvmlVgpuTypeId) GetClass() (string, Return) {
 	return string(vgpuTypeClass[:clen(vgpuTypeClass)]), ret
 }
 
+// nvml.VgpuTypeGetID
```
tariq1890: @xpahos Thank you for your patience with this PR. 

Can you add a comment explaining why we are creating this wrapper method for something that isn't a NVML C symbol? This is so we don't forget that we made an exception here.

xpahos: @tariq1890 fixed it. Is that explanation clear enough?

tariq1890: yes, thank you. resolving comment.

tariq1890:
@xpahos Almost there :). Can you ensure that your commits are signed? They need to show up as `Verified` when I click on your commit.

xpahos:
> @xpahos Almost there :). Can you ensure that your commits are signed? They need to show up as `Verified` when I click on your commit.

@tabo Sorry, I didn’t notice the bit about the certificate. Done.

tariq1890 (approved):

PR #179, opened 2026-05-06
update go-nvml with changes from cuda 13.2.1

tariq1890:
The following commands were run to update the go-nvml API with the latest changes from CUDA 13.2.1

- `make update-nvml-h`
- `make bindings`
- `go generate ./...`


tariq1890:
Raised PR #180 to address the CI failures

klueska:
This doesn't seem to update any of versioned libs, or add any of the wrapper functions as described in the README.

tariq1890:
@klueska I have added the wrapper functions. As for the new versioned methods, I am not able to do the expected symbol remapping as the input/output struct types are also versioned.  These struct versions are also different in each of these versioned methods.

klueska:
There should be an intermediate PR for CUDA 13.1.x that lands before this one.

gen/nvml/nvml.yml
```diff
@@ -52,6 +52,7 @@ TRANSLATOR:
     const:
       - {action: accept, from: "^NVML_"}
       - {action: accept, from: "^nvml"}
+      - {action: ignore, from: "NVML_MCDM_SUPPORT"}
```
klueska, line 55: Why did this need ignoring?

examples/gpm-metrics/main.go
```diff
@@ -89,7 +89,7 @@ func collectGPMMetrics(i int) error {
 		NumMetrics: 1,
 		Sample1:    sample1,
 		Sample2:    sample2,
-		Metrics: [210]nvml.GpmMetric{
+		Metrics: [333]nvml.GpmMetric{
```
klueska, line 92: Why did this number need bumping?

gen/nvml/nvml.h
klueska, line 139: You only need to create an alias to a versioned function if it appears in this list. In this update, this list is unmodified, so there are no new versioned function aliases necessary.

Makefile
```diff
@@ -182,7 +182,7 @@ clean-bindings:
 	rm -f $(PKG_BINDINGS_DIR)/zz_generated.api.go
 
 # Update nvml.h from the NVIDIA CUDA redistributable JSON
-update-nvml-h: CUDA_VERSION := 13.0.0
+update-nvml-h: CUDA_VERSION := 13.2.1
```
klueska, line 185: We need an intermediate PR for the 13.1 change before this one. We are unfortunately behind on updating this, but that doesn't mean we should just skip a version.

gen/nvml/nvml.h
```diff
@@ -11687,12 +12173,96 @@ typedef struct
...
+} nvmlPRMCounterValue_v1_t;
+
+/**
+ * PRM Counter Structure v1
+ */
+typedef struct
+{
+    unsigned int counterId;                 //!< Counter ID, one of \ref nvmlPRMCounterId_t
+    /* Input data */
+    nvmlPRMCounterInput_v1_t inData;        //!< PRM input values
```
klueska, line 12233: Since this struct has input values, we need to make sure our wrappers take this struct as an input. There should be other examples of this in older PRs.

pkg/nvml/device.go
klueska, line 1: I didn't look closely yet, but do any of the structs used here require the NVML_STRUCT_VERSION to be set?

gen/nvml/nvml.yml
tariq1890: This variable was not being rendered properly, so it led to a compilation error in `nvml.go`. Ignoring it was the only way to unblock this

Makefile
tariq1890: Understood

gen/nvml/nvml.yml
tariq1890: This is what the constant gets rendered as  in `nvml.go`

```
const (
// NO_UNVERSIONED_FUNC_DEFS as defined in go-nvml/<predefine>:24
NO_UNVERSIONED_FUNC_DEFS = 1
// MCDM_SUPPORT as defined in nvml/nvml.h:106
[77 67 68 77 95 83 85 80 80 79 82 84]
// API_VERSION as defined in nvml/nvml.h:111
API_VERSION = 13
// API_VERSION_STR as defined in nvml/nvml.h:112
API_VERSION_STR = "13"
// VALUE_NOT_AVAILABLE as defined in nvml/nvml.h:156
VALUE_NOT_AVAILABLE = -1
```

You can see that `MCDM_SUPPORT` is mis-rendered

tariq1890:
> There should be an intermediate PR for CUDA 13.1.x that lands before this one.

#182

pkg/nvml/device.go
tariq1890: Yes, this struct :

```
type RusdSettings_v1 struct {
	Version  uint32
	PollMask uint64
}
```
I've addressed this in #182

gen/nvml/nvml.h
tariq1890: Thanks for the tip, this is good to know

examples/gpm-metrics/main.go
tariq1890: This struct's schema was [updated](https://github.com/NVIDIA/go-nvml/pull/179/changes#diff-ad3ff5c9177f2f40510389d741bcbb7870124a34f24ceeaddeea362a608ccffdR1476) in the latest nvml.h.

PR #181, opened 2026-05-12
update: When no query is required for the return, directly return success instead of propagating the error code to the upper layer.

gcx-17211270:
The goal is to prevent the upper layer from having to handle excessive exceptions.


tariq1890:
Thank you @gcx-17211270. What is the error that is thrown by the method?

tariq1890:
Closing this for now. If I hear a compelling argument for a SUCCESS status to be returned in the case of an empty output, we can reopen the PR.

PR #182, opened 2026-05-14
sync nvml.h with CUDA version 13.1.2

tariq1890:
This PR synchronises the `nvml.h` file with that of the CUDA 13.1 toolkit.

The nvml header file can be found [here](https://developer.download.nvidia.com/compute/cuda/redist/cuda_nvml_dev/linux-x86_64/cuda_nvml_dev-linux-x86_64-13.1.115-archive.tar.xz)


pkg/nvml/device.go
```diff
@@ -2125,6 +2143,14 @@ func (device nvmlDevice) ReadWritePRM_v1(buffer *PRMTLV_v1) Return {
 	return nvmlDeviceReadWritePRM_v1(device, buffer)
 }
 
+func (l *library) DeviceReadPRMCounters_v1(device Device, counterList *PRMCounterList_v1) Return {
+	return device.ReadPRMCounters_v1(counterList)
+}
+
+func (device nvmlDevice) ReadPRMCounters_v1(counterList *PRMCounterList_v1) Return {
```
tariq1890: @klueska Acknowledged that I need to rework the input params of this method. Before I do that, I think we need to convert the *PRMCounterList_v1 pointer type in C to a go slice. Does `c-for-go` support this?

pkg/nvml/device.go
```diff
@@ -3547,6 +3592,14 @@ func (device nvmlDevice) GetSramUniqueUncorrectedEccErrorCounts(errorCounts *Ecc
 	return nvmlDeviceGetSramUniqueUncorrectedEccErrorCounts(device, errorCounts)
 }
 
+func (l *library) DeviceSetRusdSettings_v1(device Device, settings RusdSettings_v1) Return {
+	return device.SetRusdSettings_v1(settings)
+}
+func (device nvmlDevice) SetRusdSettings_v1(settings RusdSettings_v1) Return {
+	settings.Version = STRUCT_VERSION(settings, 1)
```
tariq1890, line 3640: setting the `STRUCT_VERSION` as requested in the earlier PR

pkg/nvml/device.go
tariq1890: never mind. I think I got the idea of the *PRMCounterList struct now. Please let me know what you think of the latest iteration of ReadPRMCounters_v1

pkg/nvml/device.go
```diff
@@ -2125,6 +2146,41 @@ func (device nvmlDevice) ReadWritePRM_v1(buffer *PRMTLV_v1) Return {
 	return nvmlDeviceReadWritePRM_v1(device, buffer)
 }
 
+func (l *library) DeviceReadPRMCounters_v1(device Device, prmCounters []PRMCounterId, localPort int) ([]PRMCounter_v1, Return) {
```
cdesiniotis, line 2149: I prefer updating this method signature (and the others added in this PR) to align more closely with the NVML API:
```suggestion
func (l *library) DeviceReadPRMCounters_v1(device Device, prmCounterList *PRMCounterList_v1) ([]PRMCounter_v1, Return) {
```

Here are some examples of existing methods where we require users to pass in a custom data type (like `PRMCounterList_v1`):
https://github.com/NVIDIA/go-nvml/blob/6d746b3c04d2e162bc816b813acff76d1b2e0eab/pkg/nvml/device.go#L2869
https://github.com/NVIDIA/go-nvml/blob/6d746b3c04d2e162bc816b813acff76d1b2e0eab/pkg/nvml/device.go#L3441
https://github.com/NVIDIA/go-nvml/blob/6d746b3c04d2e162bc816b813acff76d1b2e0eab/pkg/nvml/device.go#L3502

The main benefit here is that if NVML were to add new fields to the v1 structs, like `nvmlPRMCounterInput_v1_t`, we could support these fields without changing the method signature (and thus breaking users).

tariq1890: So, the `PRMCounterList_v1` is just the C-style representation of an array of the `PRMCounter_v1` type

```
type PRMCounterList_v1 struct {
	NumCounters uint32
	Counters    *PRMCounter_v1
}
```

I am not entirely sure if the `PRMCounterList_v1` struct itself will change beyond this

tariq1890: Note that if we take the `PRMCounterList_v1` type as the input parameter, then the users would be expected to initialise the array and provide a pinned go pointer to the first element of that array as part of the `PRMCounterList_v1` struct they pass in as input. The wrapper abstracts out all of those bits

cdesiniotis: Got it. I think in this case we have a strong argument to abstract away this complexity then. We can keep the method signature as currently implemented.

pkg/nvml/device.go
```diff
@@ -3512,6 +3578,25 @@ func (device nvmlDevice) WorkloadPowerProfileClearRequestedProfiles(requestedPro
 	return nvmlDeviceWorkloadPowerProfileClearRequestedProfiles(device, requestedProfiles)
 }
 
+// nvml.DeviceWorkloadPowerProfileUpdateProfiles_v1
+func (l *library) DeviceWorkloadPowerProfileUpdateProfiles_v1(device Device, operation PowerProfileOperation, profileTypes []PowerProfileType) Return {
+	return device.WorkloadPowerProfileUpdateProfiles_v1(operation, profileTypes)
+}
+
+func (device nvmlDevice) WorkloadPowerProfileUpdateProfiles_v1(operation PowerProfileOperation, profileTypes []PowerProfileType) Return {
```
cdesiniotis, line 3586: Thoughts on updating this method signature?

```suggestion
func (device nvmlDevice) WorkloadPowerProfileUpdateProfiles_v1(updateProfiles *WorkloadPowerProfilesUpdateProfiles_v1) Return {
```

Note, there appears to be a similar Go-wrapper method that does align more closely with the NVML method signature: https://github.com/NVIDIA/go-nvml/blob/6d746b3c04d2e162bc816b813acff76d1b2e0eab/pkg/nvml/device.go#L3502

tariq1890: Sure, will make the change

tariq1890: So, if we make this change, it would mean that users of this method would have to do the Int32 slice to bitmask conversion themselves. 

Right now, the wrapper method calls this helper which will construct the bit mask to apply the desired power profiles

```
func int32SliceToMask255(s []int32) Mask255 {
	var m Mask255
	for _, p := range s {
		if p < 0 || p >= 255 {
			continue
		}
		m.Mask[p/32] |= 1 << (uint32(p) % 32)
	}
	return m
}
```

cdesiniotis (approved):

PR #204, opened 2026-08-26
device: guard zero instance count in GetGpuInstances/GetComputeInstances

EylonKrause:
## Problem

`Device.GetGpuInstances` and `Device.GetComputeInstances` read the instance count from the profile info and immediately take the address of the first slice element:

```go
var count = info.InstanceCount
gpuInstances := make([]nvmlGpuInstance, count)
ret := nvmlDeviceGetGpuInstances(device, info.Id, &gpuInstances[0], &count)
```

When `count` is 0 (a zero-valued or hand-built `GpuInstanceProfileInfo` / `ComputeInstanceProfileInfo`, i.e. an invalid profile, since `InstanceCount` is the profile's maximum instance count) `&gpuInstances[0]` panics with an index-out-of-range instead of returning a `Return` code.

## Fix

Guard the zero case (mirroring the existing `nil`-info guard just above): return `nil, ERROR_INVALID_ARGUMENT` when the count is zero, before indexing.

## Testing

`go build ./...` and `go vet ./pkg/nvml/` pass.

### DIFF
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -2376,6 +2376,9 @@ func (device nvmlDevice) GetGpuInstances(info *GpuInstanceProfileInfo) ([]GpuIns
 		return nil, ERROR_INVALID_ARGUMENT
 	}
 	var count = info.InstanceCount
+	if count == 0 {
+		return nil, ERROR_INVALID_ARGUMENT
+	}
 	gpuInstances := make([]nvmlGpuInstance, count)
 	ret := nvmlDeviceGetGpuInstances(device, info.Id, &gpuInstances[0], &count)
 	return convertSlice[nvmlGpuInstance, GpuInstance](gpuInstances[:count]), ret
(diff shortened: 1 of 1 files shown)


myeolenv:

pkg/nvml/device.go
```diff
@@ -2376,6 +2376,9 @@ func (device nvmlDevice) GetGpuInstances(info *GpuInstanceProfileInfo) ([]GpuIns
 		return nil, ERROR_INVALID_ARGUMENT
 	}
 	var count = info.InstanceCount
+	if count == 0 {
```
kvalliyurnatt, line 2379: IIUC InstanceCount refers to the maximum instances for the profile, so if this is zero that would actually be an invalid input right ? shouldn't we be returning `ERROR_INVALID_ARGUMENT` ?

EylonKrause: Agreed — `InstanceCount` is the profile's maximum instance count, so a zero means a bogus or uninitialized profile struct rather than "no instances". Changed both guards to return `nil, ERROR_INVALID_ARGUMENT`, the same as the nil-info check right above them, and amended the commit so the message says so.

kvalliyurnatt (approved):

tariq1890 (approved):

PR #210, opened 2026-09-11
device: reject an empty slice in GetFieldValues/ClearFieldValues

EylonKrause:
## Problem

`Device.GetFieldValues` and `Device.ClearFieldValues` take the caller's `[]FieldValue` and immediately take the address of its first element:

```go
valuesCount := len(values)
return nvmlDeviceGetFieldValues(device, int32(valuesCount), &values[0])
```

With an empty slice `&values[0]` panics with an index-out-of-range instead of returning a `Return` code, so a caller that builds its field list dynamically (and ends up with none) crashes rather than getting an error it can handle.

## Fix

Return `ERROR_INVALID_ARGUMENT` for an empty slice, matching how the package already treats this case elsewhere: the caller-provided `prmCounters` slice is guarded with `if len(prmCounters) == 0 { return nil, ERROR_INVALID_ARGUMENT }`, and the zero instance-count guards added in #204 do the same.

No behaviour change for non-empty input.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/nvml/device.go
+++ b/pkg/nvml/device.go
@@ -1945,6 +1945,9 @@ func (l *library) DeviceGetFieldValues(device Device, values []FieldValue) Retur
 
 func (device nvmlDevice) GetFieldValues(values []FieldValue) Return {
 	valuesCount := len(values)
+	if valuesCount == 0 {
+		return ERROR_INVALID_ARGUMENT
+	}
 	return nvmlDeviceGetFieldValues(device, int32(valuesCount), &values[0])
 }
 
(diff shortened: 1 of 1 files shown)


rajathagasthya:
Thanks @EylonKrause! We'll take a look at this.

tariq1890 (approved):

tariq1890:
@EylonKrause  Please ensure commits are signed with GPG key (or similar). https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

EylonKrause:
Re-signed the commit — it should now show as Verified. Thanks for the review, @tariq1890.

tariq1890 (approved):

PR #212, opened 2026-09-20
Trim sample and topology results to the returned count

MaxFreedomPollard:
`GetSamples`, `GetTopologyNearestGpus`, and `SystemGetTopologyGpuSet` allocate from an initial count query, but NVML can return fewer entries on the second call. The wrappers currently expose the unused tail as zero-valued samples or device handles.

Trim each slice to the count returned on a successful second call. Error returns retain their existing behavior, including `ERROR_INSUFFICIENT_SIZE` with an increased required count. This follows the returned-count handling already used for compute-instance placements.

Added table-driven regression coverage for unchanged, reduced, and zero counts, plus initial-query errors and a growing count on failure. The six reduced/zero-count cases fail before the fix.

Validation: `make test` passes; `golangci-lint run ./pkg/...` passes. Full `make golangci-lint` reports the same existing SA4023 findings in `examples/dl/main.go` on both the base commit and this branch. Tests use NVML stubs and do not require a GPU.


pkg/nvml/device.go
```diff
@@ -262,15 +262,18 @@ func (l *library) DeviceGetTopologyNearestGpus(device Device, level GpuTopologyL
 
 func (device nvmlDevice) GetTopologyNearestGpus(level GpuTopologyLevel) ([]Device, Return) {
 	var count uint32
-	ret := nvmlDeviceGetTopologyNearestGpus(device, level, &count, nil)
+	ret := nvmlDeviceGetTopologyNearestGpusStub(device, level, &count, nil)
```
tariq1890: Can we revert the changes that involve method stubs? Those belong in a dedicated PR and where we can discuss them in detail.

